PR #6, opened 2016-12-17
GKE sql proxy

ceizner:
If customer want to use GKE CloudSQL v2.0 they will need to use CloudSQL proxy
instead of our MySQL pod.


markyjackson-taulia:
I think this can be closed @mbreitung

PR #8, opened 2017-03-30
Documentation update

scyth:


README.md
```diff
@@ -117,7 +117,15 @@ kubectl set image deployment/sysdigcloud-collector collector=quay.io/sysdig/sysd
 kubectl set image deployment/sysdigcloud-worker worker=quay.io/sysdig/sysdigcloud-backend:353 --namespace sysdigcloud
 ```
 
-Assuming the deployments have more than one replica each, the upgrade process will not cause any downtime.
+Assuming the deployments have more than one replica each, these commands will trigger rolling update process for each component. If you have configured container resource limits and you do not have spare resources in your cluster, before you update the image version, you should scale the replicas down by 1. For example, if you have 5 API component replicas running:
```
ltagliamonte: This looks more a tip than something we should add to the official doc. @adityakinifr what do you think about?

sysdigcloud/sdc-api.yaml
```diff
@@ -35,6 +35,21 @@ spec:
                 configMapKeyRef:
                   name: sysdigcloud-config
                   key: sysdigcloud.jvm.options
+          # 
+          # If you're running multiple replicas, it is advisable to configure the limits below. Limiting resources 
+          # can help stabilize rolling updates later down the road.
+          #
+
+          # resources:
```
ltagliamonte: doesn't look like a valid YAML file removing only the `#` char @scyth can you please fix this?

sysdigcloud/sdc-api.yaml
```diff
@@ -50,4 +65,4 @@ spec:
             name: sysdigcloud-config
         - name: ssl
           secret:
-            secretName: sysdigcloud-ssl-secret
\ No newline at end of file
+            secretName: sysdigcloud-ssl-secret
```
ltagliamonte: @scyth can you please also remove the `\n` at the end of the file, just trying to keep an unique style in the repo

sysdigcloud/sdc-collector.yaml
```diff
@@ -35,6 +35,21 @@ spec:
                 configMapKeyRef:
                   name: sysdigcloud-config
                   key: sysdigcloud.jvm.options
+          # 
+          # If you're running multiple replicas, it is advisable to configure the limits below. Limiting resources 
+          # can help stabilize rolling updates later down the road.
+          #
+
+          # resources:
```
ltagliamonte: @scyth same here not a valid YAML removing the only `#` char

sysdigcloud/sdc-collector.yaml
```diff
@@ -49,4 +64,4 @@ spec:
             name: sysdigcloud-config
         - name: ssl
           secret:
-            secretName: sysdigcloud-ssl-secret
\ No newline at end of file
+            secretName: sysdigcloud-ssl-secret
```
ltagliamonte: @scyth `\n` here as well

sysdigcloud/sdc-worker.yaml
```diff
@@ -50,4 +65,4 @@ spec:
             name: sysdigcloud-config
         - name: ssl
           secret:
-            secretName: sysdigcloud-ssl-secret
\ No newline at end of file
+            secretName: sysdigcloud-ssl-secret
```
ltagliamonte: @scyth `\n` here as well

sysdigcloud/sdc-worker.yaml
```diff
@@ -35,6 +35,21 @@ spec:
                 configMapKeyRef:
                   name: sysdigcloud-config
                   key: sysdigcloud.jvm.options
+          # 
+          # If you're running multiple replicas, it is advisable to configure the limits below. Limiting resources 
+          # can help stabilize rolling updates later down the road.
+          #
+
+          # resources:
```
ltagliamonte: @scyth not a YAML file removing the only`#` also here

README.md
markyjackson-taulia: @mbreitung This looks like it can be closed

PR #13, opened 2017-04-24
Include mention of clock sync via NTP

philrz:
A recent issue with an on-prem install turned out to be due to clock skew between Collector/Worker hosts. We also know that Cassandra is dependent on synchronized clocks. Therefore, adding to the README a general recommendation of NTP clock sync for all hosts in general.

cc: @ltagliamonte

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 - Running Kubernetes cluster, Kubernetes version >= 1.3.X (this guide has been tested with Kubernetes 1.3.6)
 - Sysdig Cloud quay.io pull secret
 - Sysdig Cloud license
+- Host clocks should be kept synchronized via [NTP](http://www.ntp.org/)
 
 ### Infrastructure Overview
 


ltagliamonte:
@philrz thank for the PR but I think that it's better to add this info in a `tips & tricks` section in the repo wiki.
Please open a PR there.

philrz:
I don't understand that. As we've learned, synchronized clocks is neither a tip nor a trick, but a requirement. I suppose we could add a prominent message to the README pointing people at the Wiki for additional requirements, but I don't see how that would be an improvement over just adding this one-liner to the existing list of requirements. And having a separate Wiki with no pointer to it would make it easy to miss, and we'd be left with the same Support exposure we have today.

ltagliamonte:
@philrz the fact is that have the NTP daemon properly configured and running on your servers is something so obvious that is you don't have it, what happened with one of our on-prem customers it's just a very rare misconfiguration bug.. that's why I'm a bit contrary to add it as a requirement.
It is like tell customers that to access a server using ssh they need a login/password and/or a key...
@philrz do you see my point?

philrz:
I do see your point, but it doesn't change my thoughts about how to address it. Using your example, if we ended up burning a lot of Support cycles due to customers not understanding the need for login/password/key when using SSH, I'd make a similar argument that we should mention it explicitly in our docs as well, despite the fact that it seems so trivially obvious to us.

PR #23, opened 2018-02-21
Sending to SNS notification channel from on-prem.

Chobicus:
AWS credentials are stored as kubernetes secrets and loaded as optional environment variables.
README is updated with the necessary steps.
Tested on local minikube.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,23 @@ Create the pull secret object using kubectl:
 ```
 kubectl create -f sysdigcloud/pull-secret.yaml --namespace sysdigcloud
 ```
+### Step 4: (Optional) AWS credentials
 
-### Step 4: SSL certificates
+AWS credentials must be set as environment variables if you want to publish to AWS SNS notification channel.
+
+Prerequisite:
+* AWS user created (Create access key needed for publishing (ACCESS_KEY_ID, SECRET_KEY))
+* SNS topic created
+* allow user to publish to SNS (SNS dashboard->Topics->Select topic->Other topic actions->Edit topic policy->Only these AWS users... insert user ARN, e.g. arn:aws:iam::123123123:user/Test-User)
+* Create the secret objects using kubectl:
+
+```
+kubectl create secret generic aws-key-id --from-literal=AWS_ACCESS_KEY_ID=ACCESS_KEY_ID --namespace=sysdigcloud
+kubectl create secret generic aws-key-secret --from-literal=AWS_SECRET_ACCESS_KEY=SECRET_KEY --namespace=sysdigcloud
+```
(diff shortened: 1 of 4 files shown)


ltagliamonte:
@Chobicus thank you for the PR but I think the approach is wrong.
The setting panel already allows the customer to import AWS credentials, so it would be better to use that credentials (customizing the grants example) to send SNS notification instead of asking the customer to import 2 sets of credentials.

adityakinifr:
If we do that we would need to show that in some way to the user. As an example a customer might not want to import AWS metadata or cloudwatch data but want SNS notifications.
Also given a customer could have several AWS keys, we would need to add support to designate one of those as the key to use.

ltagliamonte:
@adityakinifr in the setting page, when you click on the ADD button we display a documentation [link](https://support.sysdig.com/hc/en-us/articles/204711299) :
screen shot 2018-02-21 at 09 00 24

We need just to update the documentation. The backend already tries all the keys when fetch AWS metadata, we just need to do the same for SNS.

adityakinifr:
Yes but what about people who do not want AWS metadata or cloudwatch. I admit that might be a corner case, but I wouldn't go into the Providers page to do something with SNS.

ltagliamonte:
@adityakinifr they can just give the grants to send SNS:
```
{
    "Action": [
        "sns:Publish"
    ],
    "Effect": "Allow",
    "Resource": [
        "*"
    ]
}
```

jaykeshur:
my 2c: 
providing aws creds as env variables over the JVM args is definitely preferred.
The issue with the way that @ltagliamonte suggested would be, you would be able to configure lots of access creds on the settings page, but only 1 set of creds would be used to publish...how would a set of keys be "elected"?

ltagliamonte:
@jaykeshur the backend app already has a try all configured keys logics, so there is no "election" problem.

ltagliamonte:
@jaykeshur the solution here is not injecting the credentials as JVM parameters but exporting them as env variables.

Chobicus:
I thought that this would be a minimal change, both on coding and customer side, that would work.

Also, this change conforms with how [send to SNS works on SaaS.](https://support.sysdig.com/hc/en-us/articles/205375795-Setting-Up-Notification-Channels)

If we agree that AWS keys entered through UI should be used, I think it would require some UI changes to distinguish 2 use cases (getting the CW metrics and sending to SNS). Also, backend change is needed.

ltagliamonte:
@Chobicus I really don't think we need UI changes (apart from the user guide), and about the backend change, it is really really a small change (we are talking of a for loop over the SNS API call in case of onprem).

Chobicus:
Moving discussion to bug-tracker.

PR #29, opened 2018-04-24
Adding k8s statefulset note about ver

markyjackson-taulia:
I think it is important that we add this note


markyjackson-taulia:
@mbreitung @yofti Please review and merge. We will need this note regarding the k8s vers and statefulsets

PR #44, opened 2018-06-19
Remove storageClassName from VolumeClaimTemplates

henridf:
We don't know what storageClassName(s) will be available, so just go with the default.

### DIFF
--- a/datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
+++ b/datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
@@ -63,4 +63,3 @@ spec:
             resources:
               requests:
                 storage: 50Gi
-            storageClassName: standard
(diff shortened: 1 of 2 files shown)


markyjackson-taulia:
@henridf Can you run a quick test to verify

henridf:
Verified that it works on a kops-created cluster that has the following storageclasses (and where the original manifests didn't work):

$ kubectl get sc
```
NAME            PROVISIONER             AGE
default         kubernetes.io/aws-ebs   5h
gp2 (default)   kubernetes.io/aws-ebs   5h
```

markyjackson-taulia:
I ran a test against the on-prem and it failed, the failure is: Warning  FailedScheduling  49s (x17 over 5m)  default-scheduler  pod has unbound PersistentVolumeClaims

This is because the PVT does not have a storage class defined, as seen here for the pvc:
NAME                           STATUS    VOLUME    CAPACITY   ACCESS MODES   STORAGECLASS   AGE
data-sysdigcloud-cassandra-0   Pending                                                      19m

So what I suggest is we leave the standard in and make a note in the readme that lets the user know why we have standard and how they can change it if their storage class is different. I will let @mbreitung also chime in

markyjackson-taulia:
Closing this PR based on our PR

PR #66, opened 2018-09-13
DEVOPS-455 Add a custom Cassandra snitch

bennythejudge:
- This PR adds 4 entries used to add a custom snitch to Cassandra
- the snitch is based on the `GossipingPropertyFileSnitch`
- it uses the real hostname of the worker node obtained mounting the host `/etc/hostname` in the container
- it assumes that the hostname contains the zone of the host
- a filter is passed from the config file to extract the zone from the hostname (typically an `awk` or `cut` command)
- it can be used in IBM Cloud IKS


sysdigcloud/config.yaml
```diff
@@ -42,6 +42,37 @@ data:
...
+  # cassandra.snitch.hostname.file - the content of the file is piped through the command contained
+  # in cassandra.snitch.extract_cmd to extract the zone.
+  # The file could be the host /etc/hostname mounted in the container under /host/etc/hostname
+  # This configuration was devised for IBM IKS to get the real hostname of the worker node
+  # Example:
+  # cassandra.endpoint.snitch: "customHostBasedGossipingPropertyFileSnitch"
+  # (the entrypoint script of Sysdig Cassandra image will translate this to "GossipingPropertyFileSnitch")
+  # cassandra.snitch.hostname.file: "/host/etc/hostname"
+  # cassandra.snitch.extract_cmd: "cut -d- -f2"
+  # assuming that hostname is: "kube-dal10-cr0d89aa68f55d4c4c8bc6d677805284c7-w1.cloud.ibm"
```
mbreitung: Please anonymize this hostname

sysdigcloud/config.yaml
```diff
@@ -42,6 +42,37 @@ data:
...
+  # rack will end up containing "dal10"
```
mbreitung: Please make this value generic

sysdigcloud/config.yaml
bennythejudge: Done

sysdigcloud/config.yaml
bennythejudge: done

PR #72, opened 2018-10-26
Added statefulset non ha

Katari1:
Added a nonha mysql that uses a statefulset


Katari1:
We had some customers who didn't need HA but wanted mysql to be stateful.  Here are my modifications to make that happen.

PR #83, opened 2018-12-10
Modify elasticsearch statefulset and service to improve clustering

aduston-sysdig:
This change coincides with proposed changes to the elasticsearch container for sysdigcloud. 

It would allow elasticsearch pods to use the headless service with the `publishNotReadyAddresses` feature enabled to detect the other nodes during startup in order to initiate clustering properly. 

This requires a change to the statefulset to enable the Parallel pod management policy so that the elasticsearch pods start simultaneously and can detect each other right away in order to negotiate clustering effectively.

### DIFF
--- a/datastores/as_kubernetes_pods/manifests/elasticsearch/elasticsearch-statefulset.yaml
+++ b/datastores/as_kubernetes_pods/manifests/elasticsearch/elasticsearch-statefulset.yaml
@@ -10,6 +10,8 @@ metadata:
 spec:
   serviceName: elasticsearch-data
   podManagementPolicy: "Parallel"
+  # The number of replicas needs to be passed to the pods by ensuring that the value of ELASTICSEARCH_GOSSIP_NODES_NUM
+  # matches the number set here.
   replicas: 3
   template:
     metadata:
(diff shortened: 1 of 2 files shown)


datastores/as_kubernetes_pods/manifests/elasticsearch/elasticsearch-statefulset.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     role: elasticsearch
 spec:
   serviceName: elasticsearch-data
+  podManagementPolicy: "Parallel"
   replicas: 3
```
mbreitung: please add a big WARNING comment about how this number needs to exactly match ELASTICSEARCH_GOSSIP_NODES_NUM

datastores/as_kubernetes_pods/manifests/elasticsearch/elasticsearch-statefulset.yaml
```diff
@@ -45,7 +46,7 @@ spec:
         - name: ELASTICSEARCH_SERVICE
           value: sysdigcloud-elasticsearch
         - name: ELASTICSEARCH_GOSSIP_NODES_NUM
```
mbreitung: please add a big WARNING comment about how this number needs to exactly match replicas count

PR #104, opened 2019-01-23
add mysql yamls that reference extradbs

alfred-landrum:
I commented out the MYSQL_EXTRADB_* section so we could ensure the release yesterday went out w/o requiring any updates to our installation instructions. This PR adds copies of the mysql deployment & statefulset yamls that do reference the extradb vars, so that we can use them in k8s with scanning installation instructions.


alfred-landrum:
closing, not needed as we'll uncomment the sections in the normal mysql deployments, PR forthcoming.

PR #120, opened 2019-02-14
New ingress work

markyjackson-taulia:


sysdigcloud/sysdigcloud-scanning-ingress.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: extensions/v1beta1
```
papajulio: This file it's necessary? What's the difference with scanning-ingress.yaml ??

sysdigcloud/sysdigcloud-api-ingress.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: extensions/v1beta1
```
papajulio: Is this file necessary? What's the difference with api-ingress.yaml ??

sysdigcloud/haproxy-ingress.yaml
```diff
@@ -0,0 +1,226 @@
+---
```
papajulio: This file can deleted right? You already split this into the yaml's in the ingress-controller folder...

sysdigcloud/ingress_controller/ingress-daemonset.yaml
```diff
@@ -0,0 +1,78 @@
...
+      labels:
+        run: haproxy-ingress
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        role: ingress-controller
+      serviceAccountName: ingress-controller
+      containers:
+        - name: haproxy-ingress
+          image: quay.io/sysdig/haproxy-ingress:0.0.1
```
papajulio: To pull from here we need a ImagePullSecret right? Shouldn't we add it in this file?

sysdigcloud/ingress_controller/ingress-deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  template:
+    metadata:
+      labels:
+        run: ingress-default-backend
+    spec:
+      imagePullSecrets:
+        - name: sysdigcloud-pull-secret
+      containers:
+        - name: ingress-default-backend
+          image: quay.io/sysdig/ingress-default-backend:0.0.1
```
papajulio: To pull from here we need a ImagePullSecret right? Shouldn't we add it in this file?

sysdigcloud/api-headless-service.yaml
```diff
@@ -2,16 +2,15 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: sysdigcloud-api
+  name: api-headless-service
```
papajulio: This change needs to be updated in all the yamls that reference this service or be reverted. Or it won't work by default

sysdigcloud/collector-headless-service.yaml
```diff
@@ -2,16 +2,15 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: sysdigcloud-collector
+  name: collector-headless-service
```
papajulio: This change needs to be updated in all the yamls that reference this service or be reverted. Or it won't work by default

papajulio:
@markyjackson-taulia another side note to the other comments, I'm missing some doc about it somewhere. It will help a lot to have some intructions to follow when applying this.

sysdigcloud/haproxy-ingress.yaml
markyjackson-taulia: Correct, I forgot to remove that and will do so now

sysdigcloud/ingress_controller/ingress-daemonset.yaml
markyjackson-taulia: You are correct

sysdigcloud/api-headless-service.yaml
markyjackson-taulia: Reverted

sysdigcloud/sysdigcloud-api-ingress.yaml
markyjackson-taulia: Dup and removed

sysdigcloud/sysdigcloud-scanning-ingress.yaml
markyjackson-taulia: Dup and removed

sysdigcloud/collector-headless-service.yaml
mbreitung: this is reverted

sysdigcloud/ingress_controller/ingress-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+          args:
+              - --reload-strategy=multibinder
+              - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
```
papajulio: The service in this yamls is named differently this line should be: ` - --default-backend-service=$(POD_NAMESPACE)/ingress-default-backend ` Without this change the daemonset doesn't start properly...

sysdigcloud/collector-headless-service.yaml
```diff
@@ -7,11 +7,11 @@ metadata:
     app: sysdigcloud
     role: collector
 spec:
-  type: LoadBalancer
+  clusterIP: None
   ports:
     - port: 6443
       name: secure-collector
-      targetPort: 6443
+      targetPort: 6000
```
papajulio: I have a doubt about this port, we are exposing here the 6000 port, shouldn't be 6443 ?? Or this is ok?

sysdigcloud/api-ingress.yaml
```diff
@@ -0,0 +1,31 @@
...
+      http:
+        paths:
+          - backend:
+              serviceName: sysdigcloud-api
+              servicePort: 443
+            path: /api
+          - backend:
+              serviceName: sysdigcloud-api
+              servicePort: 443
+            path: /ui
```
papajulio: Here we are exposing only the /ui and /api endpoints from the API, but in onprem we need to access to / and get the login page that it's serverd by the API, could be that we are re-using here the staging/prod config where the frontend it's served from other part (S3??) ??

sysdigcloud/scanning-ingress.yaml
```diff
@@ -1,27 +0,0 @@
----
```
papajulio: I think this file was valid, you delete too much this time I think :)

sysdigcloud/collector-headless-service.yaml
markyjackson-taulia: This is ok. @mbreitung and I discussed

sysdigcloud/scanning-ingress.yaml
markyjackson-taulia: opps. I added it back

sysdigcloud/collector-headless-service.yaml
papajulio: @markyjackson-taulia we forgot to add this to our PR, but this change is wrong it doesn't work if you don't leave as it was, with 6443

PR #139, opened 2019-03-06
Reverting sysdigcloud.logging.stdout to false

markyjackson-taulia:


wpalmeri:
Low priority, but what was the reason for this?

PR #141, opened 2019-03-11
[QA-570] Added haproxy ingress.class

janoe:


sysdigcloud/api-ingress-with-secure.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   annotations:
+    kubernetes.io/ingress.class: "haproxy"
```
mbreitung: please also add this line to api-ingress.yaml

janoe: Done

PR #142, opened 2019-03-12
Cassandra updates for parallel startup.

aduston-sysdig:
Cassandra:
- Updated statefulset to use parallel startup.
- Added configmap example and commented-out optional parameters for jvm configuration tuning.
- Added resource limit section (commented out).
- Some formatting/cleanup

Elasticsearch
- Update to latest build
- Added resource limits to the deployment to make it consistent w/ the statefulset config
- Some formatting/cleanup

MySQL Router
- Bump the version to the latest build
- Some formatting/cleanup


datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
```diff
@@ -93,12 +107,15 @@ spec:
                   name: sysdigcloud-config
                   key: cassandra.datacenter.name
             ##########################################
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "nodetool drain"]
```
wpalmeri: what happens if this doesn't complete in time? do we need to set `terminationGracePeriodSeconds`?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
```diff
@@ -3,8 +3,12 @@ apiVersion: apps/v1beta1
 kind: StatefulSet
 metadata:
   name: sysdigcloud-cassandra
+  labels:
+    app: sysdigcloud
+    role: cassandra
 spec:
   serviceName: sysdigcloud-cassandra
+  podManagementPolicy: Parallel
```
wpalmeri: what kind of testing have we done around this? what if someone updates annotations for example and all the pods start cycling in parallel?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-deployment.yaml
```diff
@@ -13,15 +13,25 @@ spec:
         role: cassandra
     spec:
       containers:
-        - image: quay.io/sysdig/cassandra:2.1.20.17
+        - image: quay.io/sysdig/cassandra:2.1.21.2
```
wpalmeri: unrelated to this PR: what is this Deployment for?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
aduston-sysdig: That makes sense. What would be a sensible time limit for a nodetool drain?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
aduston-sysdig: We've tested this pretty extensively. I'll send you the document I wrote on it. 

If there are some specific test cases you think we should run through let me know and i can add them.

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-deployment.yaml
aduston-sysdig: I'm not sure if my understanding is 100% accurate, but I think originally we didn't have a statefulset option for these services, and just had them defined as a deployment. I think when the statefulsets were added we continued to maintain these as an option for historical reasons.

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
wpalmeri: i think we should bump to 60 seconds. putting it in here also makes it explicit that it's a setting that can be tweaked

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-deployment.yaml
```diff
@@ -13,15 +13,25 @@ spec:
...
+          imagePullPolicy: Always
+          ## Uncomment if you want to set resource limits for Cassandra.
+          ## Make sure the limits you set are appropriate for your environment.
+          # resources:
+          #   requests:
+          #    cpu: 500m
+          #    memory: 1Gi
+          #   limits:
+          #    cpu: 1000m
+          #    memory: 2Gi
```
mbreitung: Can we uncomment these lines and update sane defaults?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
```diff
@@ -17,29 +21,39 @@ spec:
...
+      #        matchExpressions:
+      #        - key: app
+      #          operator: In
+      #          values:
+      #          - sysdigcloud
+      #        - key: role
+      #          operator: In
+      #          values:
+      #          - cassandra
+      #      topologyKey: "kubernetes.io/hostname"
```
mbreitung: Can we uncomment these lines to make the best practice be default?

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-deployment.yaml
aduston-sysdig: What would you consider to be sane defaults in this case?

mbreitung:
this is waiting for https://github.com/draios/sysdigcloud-kubernetes/pull/144 to be merged

datastores/as_kubernetes_pods/manifests/cassandra/cassandra-statefulset.yaml
```diff
@@ -16,30 +20,44 @@ spec:
...
-            #topologyKey: "kubernetes.io/hostname"
+      ## Pod anti-affinity rules ensure that cassandra pods run on separate hosts.
+      ## You may need to disable these if your number of available kubernetes hosts is less than
+      ## the replication factor for your statefulset.
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: app
```
mbreitung: We have found that choosing a selector that is not the same as the service allows for the ability to add a second statefulset to the service for maintenance like disk resizing or statefulset changes. This is accompished by using an alternative label only used for affinity tagging

PR #157, opened 2019-04-02
Enable Stats Port

Katari1:
We have a default app check for haproxy that doesn't work since we don't open up the port.  As a result there is a red X that customers ask about out of the box.  This 4 line fix will allow people to look at the haproxy stats.

### DIFF
--- a/sysdigcloud/ingress_controller/ingress-daemonset.yaml
+++ b/sysdigcloud/ingress_controller/ingress-daemonset.yaml
@@ -41,6 +41,10 @@ spec:
               hostPort: 443
               name: https
               protocol: TCP
+            - containerPort: 1936
+              hostPort: 1936
+              name: stats
+              protocol: TCP
           readinessProbe:
               failureThreshold: 3
               httpGet:


sysdigcloud/ingress_controller/ingress-daemonset.yaml
```diff
@@ -41,6 +41,10 @@ spec:
               hostPort: 443
               name: https
               protocol: TCP
+            - containerPort: 1936
+              hostPort: 1936
+              name: stats
```
wpalmeri: could you name the port `haproxystats`?

wpalmeri: does the app check use the node IP? if so could we update it to use the clusterIP so that we don't need to expose the hostPort?

mbreitung: I don't think that we should expose this port at the edge.

mbreitung:
We do not want to expose the stats port outside of the cluster. I am closing this PR

PR #182, opened 2019-05-21
Events forwarder deployment

AneethAnand:


sysdigcloud/events-forwarder-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: events-forwarder
+    spec:
+      containers:
+        - name: events-forwarder
+          image: quay.io/sysdig/events-forwarder:8d1a093-05-21-2019-01-03-53
```
mbreitung: this version looks like an internal build number. Can this be converted to semantic versioning?

AneethAnand: @mbreitung  Can you please give me an example where Semantic versioning is used?

mbreitung: We are converting to semantic versioning in the next release for all components. Previously it was simply the build number which is much cleaner than a hash code+date

AneethAnand: @mbreitung , so can you merge this change then?

mbreitung: This needs a proper version

sysdigcloud/events-forwarder-deployment.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: apps/v1
```
mbreitung: please add `---` to the top of this file

mbreitung:
Closing this stale PR

PR #190, opened 2019-06-05
Remove dependence on python kubernetes.

bjhaid:
The check_storageclass.py python code checks if the argument passed to
is an existing storageclass and returns an exit code based on the
presence or absence of the storageclass, this patch replaces this
behavior with `kubectl get storageclass -o name ${STORAGE_CLASS_NAME}`
and checks for the presence of absence of a return value, which should
provide the same behavior.

Caveat I have not been able to test this change, since I currently do
not have access to a k8s cluster but I am like 95% certain the change is
correct.


markyjackson-taulia:
Disregard

PR #191, opened 2019-06-07
Devops 1238 add es tls to templating

bob-bins:


sysdigcloud/config.yaml
```diff
@@ -103,14 +103,14 @@ data:
...
+  elasticsearch.hostname: sysdigcloud-elasticsearch
   # Optional: Elasticsearch JVM options. For heavy load environments this will override the defaults (e.g. -Xms6G -Xmx6G)
   elasticsearch.jvm.options: ""
-  # Optional: Elasticsearch JVM options. Suggested value "-Xms6G -Xmx6G"
-  elasticsearch.jvm.options: ""
-  #Username required for the Search Guard authentication
+  # Username required for the Search Guard authentication
   elasticsearch.searchguard.enabled: "false"
   elasticsearch.user: "admin"
+  # This needs to be false when using Search Guard, otherwise set to true
```
mbreitung: Could this comment be removed? We could force `verifySSL=false` when we find that `  elasticsearch.searchguard.enabled: "true"`?

mbreitung: I think that the verifySSL needs to remain for the scenario where the customer is using a self hosted ES with a real cert

bob-bins: What do you think about changing the value to `false` by default and updating the comment to say something like `You may change this to true if you are running your own secure Elasticsearch cluster and desire hostname verification`.

I don't love the idea of having an exposed configurable parameter be possibly ignored based on the value of other parameters. It's a place for future confusion

bob-bins: ended up commenting the property out, saying it should only be used if the user has their own secure ES cluster. Backend has code to set the value appropriately when Searchguard is enabled

datastores/as_kubernetes_pods/manifests/elasticsearch/README.md
```diff
@@ -1,21 +1,113 @@
...
+
+        NAMESPACE=<replace-this>
+        kubectl -n $NAMESPACE create secret generic ca-certs --from-file=<path to root-ca.pem>
+
+3. Uncomment `elasticsearch.verifySSL` in `sysdig/config.yaml` and set the value appropriately.
+
+4. Follow steps 4, 7, and 8 from [Enabling authentication on an existing cluster with standard Sysdig Elasticsearch setup](#enabling-authentication-on-an-existing-cluster-with-standard-sysdig-elasticsearch-setup).
+
+# TODO: review all content below this to ensure correctness
 # Search Guard Configuration Files
```
bob-bins: If we want to initially keep things simple, I'm questioning whether this should be an option for the customer. If so, I will need to enhance this section

bob-bins:
@mbreitung should I create another branch that does not have `configurator` changes and submit a PR to `master`?

janoe:
Is this PR still valid? I see that the new code is reverting the newer changes

bob-bins:
replaced by https://github.com/draios/sysdigcloud-kubernetes/pull/216

bob-bins:
sorry @janoe, I missed your comment!

PR #194, opened 2019-06-14
DEVOPS-1805: Build configurator images from Jenkins.

bjhaid:
This will build configurator images from Jenkins that includes all of
our configuration, which will eliminate the need for the user to clone
and build the images themselves.

Currently I plan to make this a single branch Jenkins pipeline and build
only images from a "blessed" branch, if we find need for building all
branches we can revisit this decision.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,38 @@
+pipeline {
+  agent {
+    node { label "${WORKER}" }
+  }
+  environment {
+    registryCredential = 'jenkins-artifactory'
+    dockerImage = ''
+  }
+  stages {
+    stage('Build image') {
+      steps{
+        withCredentials([string(credentialsId: 'ARTIFACTORY_URL', variable: 'ARTIFACTORY_URL')]) {
+          script {
+            dockerImage = docker.build("${ARTIFACTORY_URL}/configurator:${env.BUILD_NUMBER}", './configurator/')
+          }
+        }
+      }
+    }
+    stage('Push image') {
+      steps {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+pipeline {
+  agent {
+    node { label 'builder-backend-j8' }
+  }
+  environment {
+    registry = 'docker.internal.sysdig.com'
```
mbreitung: can this value and the following be pulled in from ENV vars in the jenkins job instead of hard coding in a public repo?

bjhaid: ah I forgot this is a public repo, I will create a jenkins env var for this and pull it in

configurator/Dockerfile
```diff
@@ -18,4 +18,6 @@ RUN apk add python py-pip jq openssl ca-certificates \
       && cp /tmp/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit/oc /usr/bin/oc-bin \
       && rm -rf openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz /tmp/openshift* linux-386
```
jpravin: Ayo thats what i was doing before and changed it. Can we add the mount code for values.yaml as well ?

bjhaid: that's going to be in the `docker run  ...`, so I expect it go be:

`docker run -v ${PWD}:/generated --rm -it ${image_name}`

where the user has `values.yaml` in their PWD. I'll will add code to move verify `/generated` is mounted and contains a `values.yaml` either in this PR or a subsequent PR

Jenkinsfile
bjhaid: Addressed

PR #195, opened 2019-06-14
DEVOPS-1805 Simply steps to run configurator.

bjhaid:
This relies on only the presence of a values.yaml in current working
directory and leverages the configs add to the image to generate
manifests. This technically cuts away 2 steps from the steps required to
run the configurator, since there is no need for the user to clone repo
and build the image themselves.

Anticipated usage:

`docker run -v ${PWD}:/manifests --rm -it ${img}`

Depends on:

#194


bjhaid:
Only this commit:

https://github.com/draios/sysdigcloud-kubernetes/pull/195/commits/a6e355e0f0c77e48013681e6a602911deb65caed

belongs to this PR, the other is from 

https://github.com/draios/sysdigcloud-kubernetes/pull/194

which is yet to be merged

Jenkinsfile
```diff
@@ -0,0 +1,38 @@
+pipeline {
+  agent {
+    node { label 'builder-backend-j8' }
```
markyjackson-taulia: Can you remove this and make it an env var? We can set these internally if need be

markyjackson-taulia: You can make it `${WORKER}`

bjhaid: Thanks addressed

PR #200, opened 2019-06-17
Fix helm templates and values.yaml.

bjhaid:
`default` function in helm only works when the top level key is absent,
for hierarchical data, the hierarchy must be present and `default` will
fill in if the value is absent.

This also addressed lookups that had missing defaults.

This was an oversight I did not test after commenting out the fields, as
it currently is the fields cannot be commented out otherwise things will
break.

### DIFF
--- a/configurator/sysdig-chart/values.yaml
+++ b/configurator/sysdig-chart/values.yaml
@@ -17,12 +17,13 @@ quaypullsecret: <Insert QuaySecret>
 storageClassName: <Insert StorageClassName>
 #This drives multi-az awareness, if sysdigcloud is deployed to a multi-az setup
 #this should be updated as adequate
-# cloudProvider:
-#   defaults to an empty string
-#   name: aws
-#   defaults to false
-#   isMultiAZ: true
-#   region: us-east-1
+cloudProvider:
+  # example aws
+  name:
+  # boolean true|false
+  isMultiAZ: false
+  # example us-east-1
+  region: ""
 #supports aws | gke
 storageClassProvisioner: aws
 #supports openshift | kubernetes defaults to kubernetes install
--- a/configurator/sysdig-chart/templates/common-config/config.yaml
+++ b/configurator/sysdig-chart/templates/common-config/config.yaml
@@ -80,11 +80,11 @@ data:
   # This conditional has to be kept here because golang templates do not
   # support nesting if conditionals, it expects an if conditional to close at an
   # end statement.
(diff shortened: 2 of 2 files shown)


configurator/sysdig-chart/values.yaml
```diff
@@ -17,14 +17,14 @@ quaypullsecret: <Insert QuaySecret>
...
-#   isMultiAZ: true
-#   region: us-east-1
+cloudProvider:
+  # example aws
+  name:
+  # boolean true|false
+  isMultiAZ: false
+  # example us-east-1
+  region: ""
 #supports aws | gke
```
bob-bins: this comment refers to the line below being deleted, so it should be deleted as well

bjhaid: oh I should not be deleting that line thanks for catching it :)

bjhaid: I did not add it in my `git add` correctly :) fixed

bob-bins:
What is the benefit to adding default values to the helm template if we have those same values as the default in values.yaml? Is it reasonable to assume the user will always have a properly-formatted values.yaml file (that we provide them)?

bjhaid:
> What is the benefit to adding default values to the helm template if we have those same values as the default in values.yaml? Is it reasonable to assume the user will always have a properly-formatted values.yaml file (that we provide them)?

Good question, we can just default to an empty string and not need to use helm's `default` function, but I am beginning to think we should not use hierarchical data at all and instead just have it flat then commented out, that's going to be less error-prone.... I'll like other folks to weigh in on the later piece of my statement. @jpravin hinted towards doing something similar.

jpravin:
https://helm.sh/docs/chart_best_practices/#flat-or-nested-values

As per helm best practices. They want us to use flat.

bob-bins:
The best practices recommends flat most of the time, but it mentions there are times where nested is better. I think this is something we can look at case-by-case (from the user's perspective) rather than impose a flat-only rule.

markyjackson-taulia:
Does this add any value to the conversation: https://github.com/helm/helm/blob/master/docs/chart_best_practices/values.md

bob-bins:
That's the github version of the link Pravin posted :P

markyjackson-taulia:
Then that solves it 🤔

jpravin:
True. I agree readability vs usability. I wish there was a way yaml can interpret config that way.

bjhaid:
Wearing a users hat for a minute.... I may want to comment out/delete fields I do not care about, with flat fields this would just work fine, for nested fields errors will start popping out left, right and center, and it's not always immediately clear what to do to make this errors go away, from a UX perspective I do not think this is a great user experience.

PR #201, opened 2019-06-18
Adding missed config

jpravin:
This config has a reference in cassandra statefulset.

Am seeing this error in configurator.

Error: Couldn't find key cassandra.snitch.hostname.file in ConfigMap sysdigcloud/sysdigcloud-config


bjhaid:
thanks for catching this!

bjhaid:
Merging cause I need this change :)

PR #202, opened 2019-06-18
Support local path

bjhaid:
This PR contains 2 commits:

Cleanup yq usage. https://github.com/draios/sysdigcloud-kubernetes/commit/71986f860c7b5ac04bffe11520ec4fe2a20eec2f

and:

DEVOPS-1805 Support local path via k8s hostPath volumes. (https://github.com/draios/sysdigcloud-kubernetes/commit/5e62fc53acb860474fe2c5c73c580f713acd66c8)
This does a couple of things:

- Introduces one more option to the union field for
storageClassProvisioner `hostPath`. I avoided naming this `local`
because we might end up needing to support the
[`local`](https://kubernetes.io/docs/concepts/storage/storage-classes/#local)
provisioner.
- cleans up verifying the existence of the `STORAGE_CLASS` in
generate_templates.sh, this was failing when the storage class does not
exist and because we set the `-e`(errexit) flag this causes the script
to fail when the storage class does not exist, the workaround is to pipe
stderr to `/dev/null` and or the command to `/bin/true` preventing the
early exit.
- Moved volumeClaimTemplates towards the bottom in a couple of places to
make the conditional easier to write.


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -70,7 +70,11 @@ else
 fi
 
 echo "step5a: generate storage"
+if [[ "$(yq -r .storageClassProvisioner ${TEMPLATE_DIR}/values.yaml)" == "hostPath" ]]; then
+  echo "hostPath mode, skipping generating storage configs"
+else
 kustomize build $MANIFESTS_TEMPLATE_BASE/storage/                                      > $GENERATED_DIR/storage-class.yaml
```
mbreitung: this line needs an indent. Also the generated file output with the extra whitespace looks odd

configurator/sysdig-chart/monitor.sh
```diff
@@ -4,21 +4,25 @@ set -euo pipefail
...
-if [[ $? == 0 ]]; then
-  broadcast 'g' "StorageClass ${STORAGE_CLASS_NAME} exits. Skipping storageClass creation..."
+if [[ "$(yq -r .storageClassProvisioner /sysdig-chart/values.yaml)" == "hostPath" ]]; then
+  broadcast 'g' "hostPath mode, skipping StorageClass"
 else
-  broadcast 'g' "Creating StorageClass"
-  kubectl apply -f /manifests/generated/storage-class.yaml
+  STORAGE_CLASS_NAME=$(yq -r .storageClassName /sysdig-chart/values.yaml)
+  #Create config
+  STORAGE_CLASS="$(kubectl get storageclass ${STORAGE_CLASS_NAME} 2> /dev/null || /bin/true)"
```
mbreitung: while prepending `/bin/` to `true` is correct, we don't do that anywhere else

configurator/sysdig-chart/templates/data-stores/postgres/postgres-statefulset.yaml
```diff
@@ -85,3 +76,20 @@ spec:
         - name: data
```
mbreitung: Can multiple volume mounts have the same name? `data`?

configurator/sysdig-chart/generate_templates.sh
bjhaid: > this line needs an indent.

Thanks for catching this! Fixed

> Also the generated file output with the extra whitespace looks odd

if you switch to the unified view you will not see that, but I agree about it looking weird

configurator/sysdig-chart/monitor.sh
bjhaid: I can update... though we tend to use `true`(or `false`) in a bunch of places as the boolean literal, prefixing with `/bin` at least makes it obvious that it's the binary and not the literal, what do you think?

configurator/sysdig-chart/templates/data-stores/postgres/postgres-statefulset.yaml
bjhaid: it is a single volume mount, in the hostPath case it points at the hostPath volume otherwise it points at the volume from the volumeClaimTemplate

configurator/sysdig-chart/templates/data-stores/cassandra/cassandra-statefulset.yaml
```diff
@@ -152,6 +152,12 @@ spec:
       # - name: host-etc-hostname
       #   hostPath:
       #     path: /etc/hostname
+{{ if eq .Values.storageClassProvisioner "hostPath" }}
```
bob-bins: very minor, but it would be nice to have a `-` in some of these templating functions to trim the extra new line when the yaml is generated from this template.

E.g.
```
{{- if eq .Values.storageClassProvisioner "hostPath" }}
   ...content
{{- else }}
   ..content
{{- end }}
```

bjhaid: thanks! I'll go add that

configurator/sysdig-chart/templates/data-stores/overlays/cassandra/small/volume-claim.yaml
```diff
@@ -3,6 +3,15 @@ kind: StatefulSet
 metadata:
   name: sysdigcloud-cassandra
 spec:
+{{ if eq .Values.storageClassProvisioner "hostPath" }}
```
bob-bins: I am a bit confused - the Statefulset already contains this block, so why is it also in this overlay file?

bjhaid: I spoke to @jpravin about this, this should be cleaned up probably in a subsequent PR/commit, there are bunch of places where we have things in the overlay and in the actual config

bob-bins: I'd make the case that it's best to clean it up in this PR, since the it's being introduced in this PR!

bjhaid: this is just adding a conditional around the volumeClaims in both the statefulset config and the overlay.... and not introducing the duplication, it's probably amplifying it... I think cleaning up the duplication should be a completely separate body of work. You could suggest I should not amplify, but while the duplication is not cleaned up having conditional in one place and not the other will be confusing to whoever reads the code.

bob-bins: I disagree that having it only in one place will be more confusing - it's just following the DRY principle. But if you're adamant about keeping it in this PR I will concede! :)

configurator/sysdig-chart/values.yaml
```diff
@@ -24,7 +24,9 @@ cloudProvider:
   isMultiAZ: false
   # example us-east-1
   region: ""
-#supports aws | gke
+storageClassProvisioner: aws
```
bob-bins: you can remove this addition since it's already existing below

configurator/sysdig-chart/templates/data-stores/overlays/cassandra/small/volume-claim.yaml
bjhaid: I removed it in:

https://github.com/draios/sysdigcloud-kubernetes/pull/202/commits/050fa9da484ae6cb70a794cef75c8429d89af714

configurator/sysdig-chart/values.yaml
```diff
@@ -24,7 +24,9 @@ cloudProvider:
...
+#supports aws | gke | hostPath
```
bob-bins: this comment says it supports these 3 values, but the code only deals with two scenarios: `hostpath` or `not hostpath`

bjhaid: well aws and gke are supported and have been before this change, the only thing this change is introducing is using `hostPath` in lieu of aws or gke

configurator/sysdig-chart/templates/data-stores/overlays/cassandra/small/volume-claim.yaml
bob-bins: ah, I see. Thanks Ayo

configurator/sysdig-chart/values.yaml
bob-bins: The comment might have been outdated before your PR then, because setting `storageClassProvisioner: asdfhjklsadjf` will have the same effect as setting `storageClassProvisioner: aws` or `storageClassProvisioner: gke`. Unless I am missing something

configurator/sysdig-chart/values.yaml
bjhaid: removed

configurator/sysdig-chart/values.yaml
bjhaid: you will get an error (some kustomize one) if you set it to something not aws, gke or hostPath we just don't do explicit checks for gke and aws. `hostPath` does not create the same type of object as you get with gke or aws and it is redundant to configure aws|gke if you are using hostPath. The only reason why I am using `storageClassProvisioner` to drive `hostPath`  now is to make it deliberate that you can not configure aws|gke when `hostPath` is configured, if we eventually decide to use the [local provisioner](https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/) the conditional will go away and configs will match what we have for aws and gke

bob-bins: `you will get an error (some kustomize one) if you set it to something not aws, gke or hostPath we just don't do explicit checks for gke and aws`

Ah, I dodn't know this. Thanks for clearing it up and for the additional explanation!

bjhaid:
> friendly reminder to update the wiki page since the tag will change from 18

Thanks, will do

> Would it be good to have a latest tag in this case?

We actually push the `latest` tag, the problem with using that in the `docker run ...` instruction is that docker will not pull the image if it had previously pulled it, hence you will not have the true latest

bob-bins:
Maybe we can change the wiki steps to do a `docker pull` before the `docker run`. It beats having to update the wiki every time a change is merged!

bjhaid:
> Maybe we can change the wiki steps to do a docker pull before the docker run.

This is reasonable, however latest is problematic for these other reasons:

- There's no clear mechanism to test different versions of the release.
- When a user reports an issue it is difficult to tell which of the releases introduced the reported, with explicit tags you can rollback to reproduce the bug.

Based on the above I'll prefer to manually update the wiki, hopefully we can automate posting the tagged versions to slack at some point.

configurator/sysdig-chart/templates/data-stores/cassandra/cassandra-statefulset.yaml
```diff
@@ -30,7 +29,7 @@ spec:
       affinity:
         podAntiAffinity:
           ## Strive to spread the pods across multiple AZs
-          {{ if (.Values.cloudProvider.isMultiAZ | default false) }}
+          {{- if (.Values.cloudProvider.isMultiAZ | default false) }}
```
jpravin: Ayo can we do a different commit for this?

configurator/sysdig-chart/templates/data-stores/mysql/mysql-cluster-statefulset.yaml
```diff
@@ -162,6 +162,13 @@ spec:
               port: 10512
       imagePullSecrets:
         - name: sysdigcloud-pull-secret
+{{- if eq .Values.storageClassProvisioner "hostPath" }}
```
jpravin: Instead of doing this here. Can we Volumes and VolumeClaimTemplates to a separate kustomize layer?

jpravin: or even split volume and volume claim template.

configurator/sysdig-chart/templates/storage/kustomization.yaml
```diff
@@ -1,6 +1,8 @@
+{{- if ne .Values.storageClassProvisioner "hostPath" }}
```
jpravin: Do we need this?

jpravin: We are doing the hostPath check in the monitor.sh. Can we not safely skip this?

bjhaid: this is just being overly defensive

bjhaid: I don't feel strongly about it, I can remove it if you do

configurator/sysdig-chart/templates/data-stores/mysql/mysql-cluster-statefulset.yaml
bjhaid: That will be a bigger change since we are not using overlays for any part of this file, I also was not sure about the rationale for not using overlays for it prior to this change

bjhaid: We spoke about this and @jpravin clarified updated

PR #203, opened 2019-06-18
DEVOPS-1805 Push latest tag.

bjhaid:
This allows us provide a generic docker image tag that our end users can
always point at while we are iterating.


jpravin:
lgtm

PR #208, opened 2019-06-20
Devops 1814 redis ha configurable

jpravin:
Renamed redis-stateful to redis-ha
Renamed redis-single to redis
Earlier used redis-single for small & redis-stateful for medium and large.
Currently using small remains the same.
For medium and large install redisHa is made optional.


configurator/sysdig-chart/monitor.sh
```diff
@@ -32,6 +33,25 @@ else
   wait_for_pods 10
 fi
 
+function checkKubectlExistsYesDeletes(){
+  IS_EXISTS="$(kubectl -n $NAMESPACE get $1 $2 2> /dev/null || /bin/true)"
```
bob-bins: this isn't necessarily a change request, but a pattern I like to use with shell scripting is to put named input parameters at the top to make things a bit more readable

For example:
```
function checkKubectlExistsYesDeletes(){
  local resourceType=$1
  local resourceName=$2

  IS_EXISTS="$(kubectl -n $NAMESPACE get $resourceType $resourceName 2> /dev/null || /bin/true)"
  ...
}
```

jpravin: Makes lots of sense. Will add it.

configurator/sysdig-chart/values.yaml
```diff
@@ -54,6 +54,9 @@ sysdig:
   anchoreWorkerReplicaCount:
   scanningApiReplicaCount:
   scanningAlertMgrReplicaCount:
+  #redisHa? set to true for highly available redis
+  #default is false
```
bob-bins: IMO, can just have a single comment: `Set to true for highly-available Redis`. The other parts of the comment are redundant

configurator/sysdig-chart/templates/data-stores/overlays/redis/medium/resources.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: sysdigcloud-redis
+spec:
+  template:
+    spec:
```
bob-bins: Is it intentional that medium and small all share the same specs for now?

jpravin: This is just a place holder. Have to set correct sizes.

configurator/sysdig-chart/values.yaml
jpravin: fair enough.

configurator/sysdig-chart/templates/data-stores/overlays/redis/medium/resources.yaml
jpravin: For now doubled it. Keeping it linear like other modules.

bob-bins:
The redis statefulsets currently follow a different labeling convention than some of the other apps. Right now they only have one label, e.g. `app: redis-primary` but others (such as the redis deployment and cassandra) have two, e.g. `app: sysdigcloud` and `role: redis`

jpravin:
I used the same name as in DataSources.
https://github.com/draios/sysdigcloud-kubernetes/blob/master/datastores/as_kubernetes_pods/manifests/redis/redis-deployment.yaml#L19
And
https://github.com/draios/sysdigcloud-kubernetes/blob/master/datastores/as_kubernetes_pods/manifests/redis/redis-primary-statefulset.yaml#L5
Changing it will create problems while supporting upgrades.

jpravin:
True. It has been that way. https://github.com/draios/sysdigcloud-kubernetes/tree/master/datastores/as_kubernetes_pods/manifests/redis. Adding/Removing a label can cause problem while applying in upgrades.

jpravin:
To be honest i am not sure why we need this app label. Seems to be a constant.

bob-bins:
It might be good for us to standardize the labels (Kubernetes has a few recommendations [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)). They are a pretty powerful feature and their usefulness is multiplied if convention is followed.

The problems you mentioned might have to do with
1. Not updating resources that target the label
2. Other people having written scripts that target the old label.

With a finished Configurator©, I don't think 1 will be an issue. 2 might be, but I checked GIthub for `app=redis-primary` and it doesn't seem like any scripts target that label.

Anyone else have thoughts on this?

jpravin:
Adding more eyes. This is a good discussion point.

configurator/sysdig-chart/templates/data-stores/overlays/redis-ha/large/resources-primary.yaml
```diff
@@ -9,8 +9,8 @@ spec:
         - name: redis-primary
           resources:
             limits:
-              cpu: "500m"
-              memory: 4Gi
+              cpu: "1"
```
mbreitung: I don't think that raising these requests and limits is correct. Even on a heavily loaded system, I see around 15% of a single core being used and 15MB of memory used

configurator/sysdig-chart/monitor.sh
```diff
@@ -32,6 +33,27 @@ else
   wait_for_pods 10
 fi
 
+function checkKubectlExistsYesDeletes(){
```
mbreitung: What does this name mean? Could the name of this function be something like `deleteResourceIfExists` or simply `delete_if_exists`? I see some function using camelCase and some using snake_case.

configurator/sysdig-chart/monitor.sh
```diff
@@ -32,6 +33,27 @@ else
...
+  local k8sResourceType=$1
```
mbreitung: why have `k8s` in this name?

configurator/sysdig-chart/monitor.sh
```diff
@@ -3,14 +3,15 @@ set -euo pipefail
 
 #Important framework functions.
 . /sysdig-chart/framework.sh
+NAMESPACE=$(yq -r .namespace /sysdig-chart/values.yaml)
 
 if [[ "$(yq -r .storageClassProvisioner /sysdig-chart/values.yaml)" == "hostPath" ]]; then
   broadcast 'g' "hostPath mode, skipping StorageClass"
```
mbreitung: I know that this was not changed in this PR, but knowing that `g` == `green` does not help this code read very well. I also don't think that it needs to be quoted

configurator/sysdig-chart/monitor.sh
```diff
@@ -32,6 +33,27 @@ else
...
+  IS_EXISTS="$(kubectl -n $NAMESPACE get $k8sResourceType $k8sResourceName 2> /dev/null || /bin/true)"
+  if [[ "$IS_EXISTS" != "" ]]; then
+    kubectl -n $NAMESPACE delete $k8sResourceType $k8sResourceName
+    broadcast 'r' "Deleting $k8sResourceType $k8sResourceName : redisHa=$IS_REDIS_HA config..."
+  fi
+}
+
+#Redis safety check
+IS_REDIS_HA=$(yq .sysdig.redisHa /sysdig-chart/values.yaml)
+if [[ $IS_REDIS_HA == true ]]; then
```
mbreitung: Does this `true` need to be in quotes?

mbreitung: ok, tested that it does not, but I think that it is interpreted as a string. Please put it in quotes

configurator/sysdig-chart/templates/data-stores/overlays/redis/medium/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../../../redis
```
mbreitung: Is there a way to reference directories so that we don't need to have the `../../`?

wpalmeri: great question and probably my biggest complaint about kustomize thus far

jpravin: This was taken from kustomize example code. I will check their best practice.

configurator/sysdig-chart/templates/data-stores/overlays/redis-ha/large/resources-primary.yaml
jpravin: 15mb ? wow I will reduce the limits.

jpravin: Mark if thats the case. Do we even need sizes ?

configurator/sysdig-chart/monitor.sh
jpravin: fixed

configurator/sysdig-chart/monitor.sh
jpravin: removed

configurator/sysdig-chart/monitor.sh
jpravin: fixed

configurator/sysdig-chart/monitor.sh
jpravin: fixed

PR #209, opened 2019-06-20
Fix path to look for user supplied cert/key.

bjhaid:
Since we have instructions to tell the users to mount their working
directory in `/manifests` it is safe to assume that the key and cert
manually provided will be in the same path hence end up in `/manifests`,
we can therefore just infer the absolute path to the cert and keys and
not rely on the user making the right guess.


configurator/sysdig-chart/values.yaml
```diff
@@ -37,7 +37,10 @@ sysdig:
   dnsName: <Insert DNS Name>
   monitorVersion: 2172
   #Advanced settings
-  #autogenerate certificate by default - generate to false and provide crt,key to use existing certs
+  #autogenerate certificate by default - generate to
+  #false and provide crt,key to use existing certs, make
+  #sure the key and cert are in your current working
+  #directory
   certificate:
     generate: true
```
jpravin: Can we move this flat convention aswell?

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -57,8 +57,8 @@ if [ "$GENERATE_CERTIFICATE" = true ]; then
   fi
   cp $GENERATED_KEY $GENERATED_CRT $MANIFESTS_TEMPLATE_BASE/common-config/certs/
 else
-  CRT_FILE=$(yq -r .sysdig.certificate.crt $TEMPLATE_DIR/values.yaml)
-  KEY_FILE=$(yq -r .sysdig.certificate.key $TEMPLATE_DIR/values.yaml)
+  CRT_FILE="/manifests/$(yq -r .sysdig.certificate.crt $TEMPLATE_DIR/values.yaml)"
```
bob-bins: can we use `$MANIFESTS` instead of `/manifests`?

configurator/sysdig-chart/values.yaml
bjhaid: I think we should take some time to go over the values.yaml and see how we can clean it up, I don't think this is great for being flat... even if it was it should be addressed in a separate commit/PR

configurator/sysdig-chart/generate_templates.sh
bjhaid: `$MANIFESTS` actually points to `/sysdig-chart`, which should be cleaned up since it's confusing, I probably should have cleaned that up when I introduced volume mounting... I'll take on cleaning it up in a different PR/commit

bob-bins: it looks like `TEMPLATE_DIR=sysdig-chart` but `MANIFESTS=/manifests` on line 24!

bjhaid: thanks!

PR #214, opened 2019-06-24
Lint shell scripts.

bjhaid:
- This introduces the use of
[shellcheck](https://github.com/koalaman/shellcheck) for linting our
shell scripts
- Fix all shellcheck warnings
- Switch from `/bin/ash` to `/bin/bash` and introduce this in the
Dockerfile.
- The jenkins job will be changed from a single branch pipeline to a
multi-branch pipeline, so this runs for every PR.


Jenkinsfile
```diff
@@ -7,7 +7,25 @@ pipeline {
     dockerImage = ''
   }
   stages {
+    stage('ShellCheck') {
+      steps {
+        script {
+          sh "docker run --rm -v " +
+          "${PWD}/configurator/sysdig-chart:/sysdig-chart koalaman/shellcheck -- " +
+          "/sysdig-chart/deploy.sh " +
```
bob-bins: would this be better as a loop that applies to files ending in `.sh`? so that nobody has to remember to update the Jenkinsfile if they add a new sh script.

bjhaid: addressed

Jenkinsfile
```diff
@@ -1,13 +1,72 @@
...
+    withCredentials([[$class: 'StringBinding', credentialsId: 'SLACK_INTEGRATION_TOKEN', variable: 'SLACK_INTEGRATION_TOKEN']]) {
+      if (nonGenericMessage) {
+        message = messageType
+      } else {
+        message = "*${messageType}*: Job ${env.JOB_NAME} build ${env.BUILD_NUMBER}"
+      }
+      message += " (<${env.BUILD_URL}|Open>)"
+      def payload = JsonOutput.toJson(
+        [
+        "channel": "${env.SLACK_CHANNEL}",
```
bob-bins: indent elements to arrays please :)

bjhaid: fixed

PR #216, opened 2019-06-24
Devops 1238 add secure es to templating

bob-bins:


configurator/sysdig-chart/templates/data-stores/elasticsearch/elasticsearch-secrets.yaml
```diff
@@ -0,0 +1,197 @@
+{{ if .Values.elasticsearch.searchguard.enabled -}}
```
bob-bins: decided it's simpler to do it this way rather than support an additional overlay option. this is certainly open for debate

configurator/Dockerfile
```diff
@@ -16,7 +16,8 @@ RUN apk add bash python py-pip jq openssl ca-certificates \
       && chmod +x /usr/bin/oc /usr/bin/oc-kubectl \
       && cp /tmp/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit/kubectl /usr/bin/oc-kubectl-bin \
       && cp /tmp/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit/oc /usr/bin/oc-bin \
-      && rm -rf openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz /tmp/openshift* linux-386
+      && rm -rf openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz /tmp/openshift* linux-386 \
+      && wget https://search.maven.org/remotecontent?filepath=com/floragunn/search-guard-tlstool/1.6/search-guard-tlstool-1.6.tar.gz && tar xzvf search-guard-tlstool-1.6.tar.gz
```
bjhaid: please `rm search-guard-tlstool-1.6.tar.gz`, or move the `rm -rf` one line after and add `search-guard-tlstool-1.6.tar.gz` to it

configurator/sysdig-chart/deploy.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
+
+DIR="$(dirname $0)"
+if [[ ! -d "$DIR" ]]; then DIR="$PWD"; fi
```
bjhaid: use:
 `DIR=${$(dirname $0):-$PWD}`
 instead of the conditional

source: http://www.tldp.org/LDP/abs/html/parameter-substitution.html

configurator/sysdig-chart/framework.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+
+DIR="$(dirname $0)"
```
bjhaid: similar comment from `deploy.sh`

configurator/sysdig-chart/framework.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+
```
bjhaid: remove the extra whitespace

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
+DIR="$(dirname $0)"
```
bjhaid: similar comment from deploy.sh

configurator/sysdig-chart/monitor.sh
```diff
@@ -1,13 +1,20 @@
 #!/bin/bash
+
+DIR="$(dirname $0)"
```
bjhaid: similar comment from deploy.sh

configurator/sysdig-chart/shared-functions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/ash
```
bjhaid: please change the shell to bash.

Suggestion for the functions defined here, they are better as variables instead since the functions are being set to variables in a bunch of other places

configurator/sysdig-chart/templates/data-stores/elasticsearch/elasticsearch-secrets.yaml
bjhaid: IMO we should always have searchguard enabled in the configurator

configurator/sysdig-chart/values.yaml
```diff
@@ -74,3 +74,13 @@ sysdig:
...
   #defaults to 6443
   collectorPort:
+elasticsearch:
+  jvmOptions: ""
+  searchguard:
+    enabled: false
+    ## Section below only needed if [searchguard.enabled == true]
+    # adminUser: admin # Do not change this value
+    # verifySSL: false
+    # adminPassword: <change-me>
```
bjhaid: move the password into secrets and make it autogenerated

configurator/sysdig-chart/values.yaml
```diff
@@ -74,3 +74,13 @@ sysdig:
   #Override for default collector port
   #defaults to 6443
   collectorPort:
+elasticsearch:
+  jvmOptions: ""
+  searchguard:
+    enabled: false
+    ## Section below only needed if [searchguard.enabled == true]
+    # adminUser: admin # Do not change this value
```
bjhaid: Having this and the remaining fields commented out will make helm unhappy

configurator/sysdig-chart/values.yaml
```diff
@@ -74,3 +74,13 @@ sysdig:
...
+    # verifySSL: false
+    # adminPassword: <change-me>
+    # readonlyPassword: <change-me>
+  # TODO: add way for users to use their own certs and keys for searchguard
```
bjhaid: I do not think we should make this configurable for users, the less knobs the better. It should be implementation detail of our infrastructure and completely opaque to our users.

configurator/sysdig-chart/values.yaml
bob-bins: This actually works fine. Helm will evaluate missing "leaf" variables as an empty string. It's only an error if the parent is missing but the leaf is referenced

configurator/sysdig-chart/values.yaml
bob-bins: The reason for this is because some customers run their own Elasticsearch cluster and we need backend to be able to authenticate with it. I agree that if they are deploying Elasticsearch using the configurator we should just have them use the certs we generate for them.

configurator/sysdig-chart/templates/data-stores/elasticsearch/elasticsearch-secrets.yaml
bob-bins: good point!

configurator/sysdig-chart/deploy.sh
bob-bins: apparently this syntax doesn't work if the value on the left is a function to be evaluated. I had to change it to `DIR="$( cd "$(dirname "$0")" ; pwd -P )"`

configurator/sysdig-chart/values.yaml
```diff
@@ -74,3 +74,13 @@ sysdig:
   #Override for default collector port
   #defaults to 6443
   collectorPort:
+elasticsearch:
+  jvmOptions: ""
+  searchguard:
+    enabled: true
+    ## Section below only needed if [searchguard.enabled == true]
+    adminUser: admin # Do not change this value
```
bob-bins: i don't like how I'm exposing a variable to the customer but then telling them to not change it. But the alternative is to have yet another values-2.yaml file that populates this variable and is read in by Helm. Not the most elegant solution but it might be preferable since it improves the user experience. Thoughts?

configurator/sysdig-chart/values.yaml
bjhaid: >  I agree that if they are deploying Elasticsearch using the configurator we should just have them use the certs we generate for them.

Since this is the configurator, we probably should not make the this configurable

configurator/sysdig-chart/defaultValues.yaml
```diff
@@ -0,0 +1,6 @@
+# Elasticsearch auth and TLS encryption
```
bob-bins: This file is meant for values that are shared among different k8s resources, but are not meant for the user to modify. If you have a better suggestion for the name of the file I'm all ears

bjhaid: One question, do you think we will change any of the values here?

configurator/sysdig-chart/templates/data-stores/cassandra/cassandra-statefulset.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
     app: sysdigcloud
     role: cassandra
 spec:
+  updateStrategy:
+    type: OnDelete
```
bjhaid: In my experience `OnDelete` has been a landmine, the person effecting the change to an `OnDelete` type pod does not experience the effect of their change, instead their change triggers when a pod is delete for other reasons. Their change could be broken but there is no way they would get the feedback and someone else has to deal with the fallout of their changes without adequate context on what is going on.

configurator/sysdig-chart/templates/data-stores/elasticsearch/elasticsearch-statefulset.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
     component: elasticsearch
     role: elasticsearch
 spec:
+  updateStrategy:
+    type: OnDelete
```
bjhaid: See my comment on cassandra yaml

configurator/sysdig-chart/defaultValues.yaml
bob-bins: I don't think the values here will change. There may be values added in the future that might change, but the purpose of this file is to have these values all in a single place rather than hard-code it in multiple places.

configurator/sysdig-chart/templates/data-stores/cassandra/cassandra-statefulset.yaml
bob-bins: The default behavior for `v1beta1` is already `OnDelete`, so there is technically no functional change here.

That being said, I agree with the cons you posted for `OnDelete`. In addition, we will need to write our own scripts for performing the update, which will largely do the same thing that a `RollingUpdate` already does. But it also gives us the freedom to customize things if we need to since we will have full control over how the pods are restarted. For example, if we enable Searchguard for Elasticsearch, a full restart of the cluster is necessary.

I don't have enough experience to have a strong opinion on which one is best to set as default though. @jpravin @Mark @markyjackson-taulia @wpalmeri, any thoughts?

bob-bins:
rebasing is all done, but the install is not coming up healthy because the monitorVersion used does not include the secure-elasticsearch changes. But 2271, which does include those changes, is not compatible with anchore v0.3.4.1

I will need a bit more time to ensure all the versions are compatible

bob-bins:
rebase is done and changes have been tested for a fresh install

PR #221, opened 2019-06-25
Introduce test harness.

bjhaid:
### What does this do:

- Adds generated configs(`sysdig.json`) for our dynamic configuration
combinations.
- Introduces a test.sh script that will attempt to regenerate the
configs and fail if the regenerated config does not match the
pre-generated configs.
- The script also takes a `config_gen` argument that will allow the user
regenerate the various `sysdig.json` files.
- Modifies the jenkins pipeline to run the tests as part of the pipeline
steps.
- Also moves the build, test and push steps into one stage that way they
are guaranteed to run on the same agent, when we get a newer jenkins we
can use [sequential
stages](https://jenkins.io/doc/book/pipeline/syntax/#sequential-stages)
instead.

All of these allows us make changes with some level of confidence, also
let the reviewer see what the effect of various changes will look like.

### Anticipated workflow:

- Make changes to various scripts
- Rebuild the image: `docker build ./configurator -t configurator`
- Test your changes:
```bash
docker run -it --rm -v \
  ${PWD}/configurator/sysdig-chart:/sysdig-chart --entrypoint \
  /sysdig-chart/test.sh docker.internal.sysdig.com/configurator
```
- If needed regenerate configs:
```bash
docker run -it --rm -v \
  ${PWD}/configurator/sysdig-chart:/sysdig-chart --entrypoint \
  /sysdig-chart/test.sh docker.internal.sysdig.com/configurator \
  config_gen
```


bjhaid:
Apologies for throwing a commit with a large number of generated files, you can ignore the `sysdig.json` files in this PR, but please do not ignore them when someone changes them in the future :)

bjhaid:
Closing this PR to create a new PR without conflicts, there's no way to fix this without force pushing or creating a merge commit

PR #222, opened 2019-06-25
Introduce test harness.

bjhaid:
### What does this do:

- Adds generated configs(`sysdig.json`) for our dynamic configuration
combinations.
- Introduces a test.sh script that will attempt to regenerate the
configs and fail if the regenerated config does not match the
pre-generated configs.
- The script also takes a `config_gen` argument that will allow the user
regenerate the various `sysdig.json` files.
- Modifies the jenkins pipeline to run the tests as part of the pipeline
steps.
- Also moves the build, test and push steps into one stage that way they
are guaranteed to run on the same agent, when we get a newer jenkins we
can use [sequential
stages](https://jenkins.io/doc/book/pipeline/syntax/#sequential-stages)
instead.

All of these allows us make changes with some level of confidence, also
let the reviewer see what the effect of various changes will look like.

### Anticipated workflow:

- Make changes to various scripts
- Rebuild the image: `docker build ./configurator -t configurator`
- Test your changes:
```bash
docker run -it --rm -v \
  ${PWD}/configurator/sysdig-chart:/sysdig-chart --entrypoint \
  /sysdig-chart/test.sh docker.internal.sysdig.com/configurator
```
- If needed regenerate configs:
```bash
docker run -it --rm -v \
  ${PWD}/configurator/sysdig-chart:/sysdig-chart --entrypoint \
  /sysdig-chart/test.sh docker.internal.sysdig.com/configurator \
  config_gen
```


bjhaid:
Apologies for throwing a commit with a large number of generated files, you can ignore the sysdig.jsonfiles in this PR, but please do not ignore them when someone changes them in the future :)

configurator/sysdig-chart/tests/resources/large_aws/values.yaml
```diff
@@ -0,0 +1,77 @@
...
+  workerReplicaCount:
+  anchoreCoreReplicaCount:
+  anchoreWorkerReplicaCount:
+  scanningApiReplicaCount:
+  scanningAlertMgrReplicaCount:
+  #SMTP Configurations
+  smtpServer: "foo.smtp.com"
+  smtpServerPort: "25"
+  #User,Password can be empty if the server does not require authentication
+  smtpUser: "foo"
```
mbreitung: The only values in a yaml that need quotes are numbers. This file has a mix of strings that are quoted and no quoted. Please be consistent.

configurator/sysdig-chart/test.sh
```diff
@@ -0,0 +1,70 @@
...
+    if [[ -d "$directory" ]]; then
+      cp "$directory/values.yaml" /sysdig-chart/values.yaml
+      /sysdig-chart/generate_templates.sh
+      TMP_FILE=$(mktemp)
+      concat_config /manifests/generated/ "$TMP_FILE"
+      sanitize "$directory/sysdig.json"
+      sanitize "$TMP_FILE"
+
+      if ! diff -w "$directory/sysdig.json" "$TMP_FILE"; then
+        broadcast 'r' "generated config does not match ${directory}sysdig.json"
```
mbreitung: Can this `'r'` be converted to `green` (without quotes)

configurator/sysdig-chart/test.sh
```diff
@@ -0,0 +1,70 @@
...
+
+  local tmp_file
+  tmp_file=$(mktemp)
+  yq -r 'select(.kind != "Secret")' "$2" > "$tmp_file"
+  mv "$tmp_file" "$2"
+}
+
+function sanitize() {
+  for field in "${IGNORED_FIELDS[@]}"; do
+    sed -i "/${field}/d" "$1"
```
mbreitung: please always use `-e` in front of all expressions for consistency

configurator/sysdig-chart/test.sh
bjhaid: we are reporting an failure here and not success, and by convention I know of red is usually used for failure, also `green` is not supported from the last commit(https://github.com/draios/sysdigcloud-kubernetes/blob/e62af8e82cff5e0e59482d40bd8d240aeb4a7e82/configurator/sysdig-chart/framework.sh#L13-L21) in the Templating_k8s_configuration branch

configurator/sysdig-chart/test.sh
bjhaid: done in:

https://github.com/draios/sysdigcloud-kubernetes/pull/222/commits/15dc88281bdc8ce801a0be5fecb50aea35321b5e

configurator/sysdig-chart/tests/resources/large_aws/values.yaml
bjhaid: done in:

https://github.com/draios/sysdigcloud-kubernetes/pull/222/commits/f1a645ea6120721a4eb26988186a17b97e50dbdf

bjhaid:
@markyjackson-taulia 

> but I wonder if we should have some type of a README or perhaps this will go in the larger documentation?

README updated in:

https://github.com/draios/sysdigcloud-kubernetes/pull/222/commits/184a2175ebaa9e54b102fc9db1fcc64320f6e43a

configurator/sysdig-chart/test.sh
```diff
@@ -0,0 +1,70 @@
...
+IGNORED_FIELDS=(
+  sysdigcloud.default.user.password
+  redis.password
+  mysql.root.password
+  mysql.password
+)
+
+# concatenates yaml files and remove objects of kind secret from the file.
+# Arguments:
+# 1 directory glob
```
bob-bins: can we give variable names to $1, $2, etc.? It would be more self-documenting than a comment

configurator/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+$> make build
+```
+
+this will build and tag a configurator docker image, to override the image/tag
+name, set the environment variable `DOCKER_REPO`, e.g:
+
+```bash
+DOCKER_REPO=my_awesome_image:my_awesome_tag make build
```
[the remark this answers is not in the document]
bjhaid: fair nit, I'll update to IMAGE_NAME instead

configurator/sysdig-chart/test.sh
bjhaid: done in:

https://github.com/draios/sysdigcloud-kubernetes/commit/1cfea8f17982aab691b808e47476d2b8d7475fc9

configurator/sysdig-chart/test.sh
```diff
@@ -0,0 +1,72 @@
...
+  for directory in /sysdig-chart/tests/resources/*/; do
+    if [[ -d "$directory" ]]; then
+      cp "$directory/values.yaml" /sysdig-chart/values.yaml
+      /sysdig-chart/generate_templates.sh
+      concat_config /manifests/generated/ "$directory/sysdig.json" "false"
+    fi
+  done
+}
+
+if [[ "${1:-"test"}" == "config_gen" ]]; then
```
bob-bins: what's the purpose of defaulting `$1` to `test` if it's only being compared against `config_gen`?

bjhaid: it has to default to something otherwise bash will complain about `unbound variable`, and I chose to default it to what the else piece of the condition is

configurator/README.md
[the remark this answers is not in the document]
bjhaid: done in:

https://github.com/draios/sysdigcloud-kubernetes/commit/2c61c0368d7686dd029819c2189c287c51f8747a

Jenkinsfile
```diff
@@ -55,63 +47,71 @@ pipeline {
...
-    stage('Build image') {
-      when {
-        branch 'Templating_k8s_configurations'
-      }
+    stage('Test') {
       steps{
         withCredentials([string(credentialsId: 'ARTIFACTORY_URL', variable: 'ARTIFACTORY_URL')]) {
           script {
-            dockerImage = docker.build("${env.ARTIFACTORY_URL}/configurator:${env.BUILD_NUMBER}", './configurator/')
+              sh(
```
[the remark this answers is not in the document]
bjhaid: we do what I have in a bunch of other places, so I'll keep it for consistency

configurator/sysdig-chart/test.sh
bob-bins: ah i see. I think `"${1:-}"` will work in this case

mbreitung: You don't need quotes around `"test"`

mbreitung: I do think that declaring variables outside of the `if` for `$1` is cleaner

configurator/sysdig-chart/test.sh
mbreitung: right, but `broadcast 'r' "g....` is not readable

bjhaid: I do not disagree, but `broadcast` is a thing that exists prior to this code and is used this way all over this repo, I'll prefer we change that everywhere in a separate PR

configurator/sysdig-chart/test.sh
bjhaid: Addressed in:

https://github.com/draios/sysdigcloud-kubernetes/commit/06c0ccfeb660924fde2edbb9d7489bb393ec0547

PR #223, opened 2019-06-25
set OnDelete explicitly as the updateStrategy for Elasticsearch and C…

bob-bins:
…assandra


wpalmeri:
If we make these onDelete, we'll need to call it out explicitly in the tool/docs and ideally provide a way to safely do the update.

bob-bins:
@wpalmeri I spoke with Yofti and he said he is not concerned about doing a no-downtime update unless it's a "very hard ask". Thus, the docs will say something like `kubectl delete po -l role=elasticsearch`. Is this something we'd want to push back on?

wpalmeri:
We will need to document this in general for any on-prem installation.

PR #229, opened 2019-06-26
TSOPS-328: Support bundle improvements.

Roland0:
3 Commits:

1) Added getting k8s node and event data
2) Added feature to get previous container logs after restart
3) Changed descriptions from JSON to YAML

### DIFF
--- a/scripts/get_support_bundle.sh
+++ b/scripts/get_support_bundle.sh
@@ -29,6 +29,11 @@ for pod in ${SYSDIGCLOUD_PODS}; do
     containers=$(kubectl ${KUBE_OPTS} get pod ${pod} -o json | jq -r '.spec.containers[].name')
     for container in ${containers}; do
         kubectl ${KUBE_OPTS} logs ${pod} -c ${container} > ${LOG_DIR}/${pod}/${container}-kubectl-logs.txt
+        restarts=$(kubectl ${KUBE_OPTS}  get pod ${pod} --no-headers | awk '{print $4}')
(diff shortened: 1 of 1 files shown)


scripts/get_support_bundle.sh
```diff
@@ -37,10 +42,17 @@ for object in svc deployment sts pvc daemonset ingress replicaset; do
...
     for item in ${items}; do
-        kubectl ${KUBE_OPTS} get ${object} ${item} -o json > ${LOG_DIR}/${object}/${item}-kubectl.json
+        kubectl ${KUBE_OPTS} get ${object} ${item} -o yaml > ${LOG_DIR}/${object}/${item}-kubectl.yaml
     done
 done
 
+# get node data, might be nice to have a selector here eg: -l role=sysdigcloud
+kubectl get nodes > ${LOG_DIR}/nodes.txt
+echo " " >> ${LOG_DIR}/nodes.txt
+kubectl describe nodes >> ${LOG_DIR}/nodes.txt
```
mbreitung: why not just push `descibe nodes` to a different file instead of appending?

Roland0: @mbreitung I thought I'd add all of the "node" data into nodes.txt

It seems to work ok, I'm open to any suggestions etc

mbreitung: How about creating a single file that has the list of nodes and a separate file for each node's `describe`?

Roland0: Good idea.

PR #230, opened 2019-06-26
Introduce support for airgap installs.

bjhaid:
What does this do:

- Requires user to mount the docker socket into the container.
- Update imagepullsecrets to reflect the credentials of the docker registry
- Does a docker login to configured registry.
- Does a docker login to quay.io using the quaypullsecret configured,
doing this is fine since the filesystem for the container is ephemeral
and user already has the credentials.
- Parses generated configs to determine docker images to pull.
- Pulls images from quay.io.
- Pushes the pre-configured images to the configured registry.


configurator/sysdig-chart/airgap.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+IMAGES=(
+  sysdig/cassandra:2.1.21.12
+  sysdig/elasticsearch:5.6.16.1
+  sysdig/mysql:5.6.43.2
+  sysdig/postgres:10.6-alpine.1
+  sysdig/redis:4.0.12.1-alpine-ha
+  sysdig/redis:4.0.12.1-alpine
+  sysdig/sysdigcloud-backend:2304-email-renderer
```
jpravin: 2304 can be from template monitorVersion

configurator/sysdig-chart/airgap.sh
```diff
@@ -0,0 +1,54 @@
...
+  sysdig/sysdigcloud-backend:2304-nginx
+  sysdig/sysdigcloud-backend:2304-allinone-java
+  sysdig/ingress-default-backend:1.0
+  sysdig/haproxy-ingress:v0.7-beta.7
+  sysdig/scanning:2304
+)
```
jpravin: missing quay.io/sysdig/anchore

configurator/sysdig-chart/airgap.sh
wpalmeri: i know this is a test to get it working, but to do this right I think we would want to parse all the yamls and pull out the `image` values

wpalmeri:
excepting my comment about building the image list dynamically, I think that this is a reasonable first pass at airgap support that will work in most cases where the installer has access to push images to an internal registry that k8s nodes can then use.

I assume the other thing to do would be to template the registry in our yamls as well

configurator/sysdig-chart/airgap.sh
bjhaid: Addressed in:

https://github.com/draios/sysdigcloud-kubernetes/commit/b8c8a103f6c1e6d1285f5182112c4fa3a6fbca9a

configurator/sysdig-chart/install.sh
```diff
@@ -31,12 +33,22 @@ do
  fi
 done
 
+if [[ "$AIRGAPPED" != "null" ]];
```
mbreitung: if putting the `then` on the next line, then the `;` is not needed. Also please choose a pattern and follow it. I see the other style `if [[ .. ]]; then` elsewhere in this set of scripts.

mbreitung: I think that I see `if [[ .. ]]; then` and `while ... ; do` as the pattern in most of our repos

bjhaid: <joke> it is consistent with the code surrounding it :)</joke> I'll fix

configurator/sysdig-chart/airgap.sh
jpravin: https://github.com/draios/sysdigcloud-kubernetes/blob/Templating_k8s_configurations/configurator/sysdig-chart/templates/sysdig-cloud/secure/anchore/anchore-core-deployment.yaml#L43

bjhaid:
I managed to add files owned by root into the git history of this branch and I have managed to confuse Jenkins, I'm going to close this PR, delete the branch to unconfuse Jenkins

PR #232, opened 2019-06-27
added flags to ignore some shellcheck checks and addressed a few shel…

bob-bins:
…lcheck codes


bob-bins:
Ignoring these two checks:
https://github.com/koalaman/shellcheck/wiki/SC1090
https://github.com/koalaman/shellcheck/wiki/SC2164

relating to sourcing `shared-values.sh`

PR #234, opened 2019-06-28
Remove dependence on kubectl get in the generate code path.

bjhaid:
`kubectl get` assumes the presence of authenticated Kubernetes
credentials which is not true when we run tests in Jenkins.

This replaces checking for the presence of ca-cert secrets in the
generate code path with checking for the presence and non-emptiness of
the generated elastic-search tls directory.


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -39,15 +39,12 @@ else
...
 log info "step3.5: creating elasticsearch certs for Searchguard"
-if kubectl -n "$K8S_NAMESPACE" get secret "$SECRET_NAME"; then
-  log info "Secret 'ca-certs' already exists. Skipping elasticsearch tls cert creation"
-else
+if [[ -z "$(ls -A "${MANIFESTS}/elasticsearch-tls-certs")" ]]; then
   log info "Generating certs for Searchguard..."
   (cd /tools/
-    ./sgtlstool.sh -c "$TEMPLATE_DIR/elasticsearch-tlsconfig.yaml" -ca -crt
-    cp -r out "${MANIFESTS}/elasticsearch-tls-certs")
+  ./sgtlstool.sh -c "$TEMPLATE_DIR/elasticsearch-tlsconfig.yaml" -ca -crt
```
mbreitung: please indent continuation lines. Also do these lines need to be in a sub-shell?

bjhaid: without the subshell, the code will need to do `pushd` `popd` or `cd -` dances

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -39,15 +39,12 @@ else
     helm template -x "templates/$SECRET_FILE" "$TEMPLATE_DIR/secret-generator" > "$GENERATED_SECRET_FILE"
 fi
 
-SECRET_NAME="ca-certs"
 log info "step3.5: creating elasticsearch certs for Searchguard"
-if kubectl -n "$K8S_NAMESPACE" get secret "$SECRET_NAME"; then
-  log info "Secret 'ca-certs' already exists. Skipping elasticsearch tls cert creation"
-else
+if [[ -z "$(ls -A "${MANIFESTS}/elasticsearch-tls-certs")" ]]; then
```
bob-bins: This change will be sufficient for the install case. When we start supporting upgrades in v2/v3 we will need to look back at this to ensure secrets are not overwritten

bjhaid: we can create the secret in a dry-run mode during code generation then apply it in the deploy phase

PR #240, opened 2019-07-03
 Introduce support for airgap installs

bjhaid:
This introduces 2 workflows for airgap installs:

- Installation machine with network access to quay and private registry
- Full airgap install, with installation machine not having access to quay

For the first workflow:
- This generates all the configs from the templates and parses them for docker images.
- It pulls all the docker images parsed and pushes them to the configured airgap registry.
- Updates imagepullsecret to point at the private registry.

For the second workflow:
- This creates a docker image containing a tar file that has all the docker images including the configurator image embedded.
- The tar file is self extracting and executable, so the user can run it by invoking `./sysdig_configurator.tar.gz`.
- The tar file extracts itself and loads the docker cache of the installation machine with all the docker images and then kicks off a docker run with the workflow now exactly similar to the first workflow.

See README for detailed workflow from user's perspective.

Couple of open concerns, naming mostly:
- I have been using *uber* for lack of better naming, please if you have suggestions I am all ears.
- I have used airgap registry and I am having second thought at calling it private registry instead, if you have preference please indicate.
- In the README I referred to a jump machine, for lack of better name, if you have a better name please suggest.
- I'll like reviewers to pay special attention to the README and give it a thorough read.


Jenkinsfile
```diff
@@ -40,7 +40,6 @@ pipeline {
   }
   environment {
     registryCredential = 'jenkins-artifactory'
-    dockerImage = ''
     SLACK_CHANNEL = '#dev-infra-updates'
```
markyjackson-taulia: Please change/remove this

configurator/Dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM openjdk:13-jdk-alpine
```
markyjackson-taulia: Are we pulling from an upstream repo? Can we maybe change this?

configurator/Dockerfile.uber
```diff
@@ -0,0 +1,3 @@
+FROM debian:buster-slim
```
markyjackson-taulia: Same question about upstream repo

configurator/README.md
```diff
@@ -45,15 +45,138 @@ The rationale for the workflow is it helps the code reviewers to see the
 effect(s) of the changes introduced, it also helps exercise the config
 generation workflow.
 
-### Usage
+## Usage
+
+### Non-Airgap deployment
+
+This assumes the Kubernetes cluster has access to pulling images from quay.io.
```
markyjackson-taulia: Perhaps it would be good to add a blurb or link to something for creating image pull secrets.

configurator/Dockerfile
mbreitung: I thought that we were moving away from alpine? in favor of debian based?

bjhaid: this was not introduced in this PR

configurator/sysdig-chart/airgap.sh
```diff
@@ -0,0 +1,126 @@
...
+  (
+    cd "$tmp_dir"
+    docker tag "${configurator_image}" configurator
+    docker save configurator -o configurator.tar.gz
+    for image in $(yq -r '.spec.template.spec | {containers: .containers[]?}.containers.image' /manifests/generated/*.yaml); do
+      img_without_path=${image##*/}
+
+      if [[ -z $(docker images -q "$image") ]]; then
+        logger info "Pulling ${image}"
+        docker pull "$image"
```
markyjackson-taulia: Does this need {}

configurator/sysdig-chart/airgap.sh
```diff
@@ -0,0 +1,126 @@
...
+      fi
+
+      logger info "Saving ${image} as ${img_without_path}.tar.gz"
+      docker save "$image" -o "${img_without_path}.tar.gz"
```
markyjackson-taulia: same here {}

configurator/sysdig-chart/uber_config/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  scanningAlertMgrReplicaCount:
+  #set to true for highly available redis
+  redisHa: false
+  #SMTP Configurations
+  smtpServer: ""
+  smtpServerPort: "25"
+  #User,Password can be empty if the server does not require authentication
+  smtpUser: ""
+  smtpPassword: ""
+  smtpProtocolTLS: "false"
```
mbreitung: in the `small_aws_airgap/values.yaml` these false are not quoted. Do they need to be?

configurator/sysdig-chart/airgap.sh
bjhaid: addressed in:

https://github.com/draios/sysdigcloud-kubernetes/pull/240/commits/788277f8a6f4a7a710bd2f4d01317d301f0a8947

configurator/README.md
bjhaid: We create the pull secrets for our users... so they do not need to create it themselves, maybe I should update the README to reflect that the cluster needs network access to talk to quay.io, so it is not confusing, what do you think?

configurator/sysdig-chart/airgap.sh
bjhaid: addressed in:

https://github.com/draios/sysdigcloud-kubernetes/pull/240/commits/788277f8a6f4a7a710bd2f4d01317d301f0a8947

configurator/README.md
bjhaid: Addressed in:

https://github.com/draios/sysdigcloud-kubernetes/commit/6f32fbdf62bd5641f0c17a2727b2ce9d629459a0

configurator/Dockerfile
markyjackson-taulia: this is moving to a larger discussion. Resolving this for now

configurator/sysdig-chart/uber_config/values.yaml
bjhaid: addressed in https://github.com/draios/sysdigcloud-kubernetes/commit/9d712d81e75f27154550673fced21cfcc8c8e5ed

PR #244, opened 2019-07-10
Local storage

bjhaid:
This leverages
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner
to create local storageClassProvisioner that setups without a dynamic
provisioner can use, the README covers pre-requisites for using the
local volume provisioner and provides an example a user can follow.

This does not deprecate hostPath storageClassProvisioner as that is
useful for the all-in-a-box setups.

This also reduces the size for small postgres storage to 30Gi to align
with other small storage.


configurator/sysdig-chart/values.yaml
```diff
@@ -1,38 +1,34 @@
 #defines which scripts needs to be run
 #generate - performs templating & kustomization
 #deploy   - apply's the generated script in k8s environment
-scripts: generate deploy
+scripts: generate
```
jpravin: I dont think this is a intended change.

configurator/sysdig-chart/tests/resources/small_aws_local/sysdig.json
```diff
@@ -1515,7 +1515,7 @@
                 }
               }
             ],
-            "image": "quay.io/sysdig/elasticsearch:5.6.16.2",
+            "image": "quay.io/sysdig/elasticsearch:5.6.16.4",
```
bob-bins: please change this to `5.6.16.5`. I introduced a breaking change that appeared in 5.6.16.4 because I didn't know at the time of the change that the Jenkins build does not build the on-prem images - they need to be built and pushed manually.

configurator/sysdig-chart/tests/resources/small_aws_local/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+#Sysdig application config
+sysdig:
+  agentCount: 10
+  #Set Sysdig license
+  license: foobar
+  dnsName: foo.bar.com
+  monitorVersion: 2304
+  anchoreVersion: v0.3.4.1
+  cassandraVersion: 2.1.21.13
+  elasticsearchVersion: 5.6.16.4
```
bob-bins: please change this to `5.6.16.5`. I introduced a breaking change that appeared in 5.6.16.4 because I didn't know at the time of the change that the Jenkins build does not build the on-prem images - they need to be built and pushed manually.

configurator/sysdig-chart/values.yaml
```diff
@@ -1,38 +1,34 @@
...
-size: medium
+size: small
 #Set k8s namespace for sysdigcloud
 namespace: sysdigcloud
 #Set Quay.Io secrets
-quaypullsecret: <change-me>
+quaypullsecret: foo
 #StorageClass config
-storageClassName: <change-me>
+storageClassName: foo
```
bob-bins: i'd prefer more descriptive placeholders than `foo`. `default` might even be ok for `storageClassName` specifically

configurator/sysdig-chart/values.yaml
```diff
@@ -1,38 +1,34 @@
...
 cloudProvider:
   # example aws
-  name: 
+  name: aws
   # boolean true|false
   isMultiAZ: false
   # example us-east-1
-  region: ""
-#supports aws | gke | hostPath
+  region:
```
bob-bins: If this parameter is intentionally optional I think we should explicitly point that out in the comment above.

configurator/sysdig-chart/values.yaml
bjhaid: the changes in this file is unintentional minus the `| local` introduction, it's a combination of how `config_gen` works and me not reading the diff after running `config_gen` to see that this file was changed

configurator/sysdig-chart/tests/resources/small_aws_local/values.yaml
bjhaid: addressed

configurator/sysdig-chart/values.yaml
bjhaid: this has been reverted

configurator/sysdig-chart/values.yaml
bjhaid: this has been reverted

configurator/sysdig-chart/values.yaml
bjhaid: this has been reverted

configurator/sysdig-chart/values.yaml
bob-bins: thanks for the context!

bjhaid: FWIW we have a ticket to fix some other problems around testing that will indirectly address the config_gen problems

configurator/README.md
```diff
@@ -185,3 +185,50 @@ https://raw.githubusercontent.com/draios/sysdigcloud-kubernetes/Templating_k8s_c
...
+      mkfs.ext4 /vol\${i}; \
+      mount /vol\${i} /sysdig/vol\${i}; \
+    done \
+  '"
+```
+
+The above creates 3 35G volumes per Kubernetes node. _Use the above as a last
+resort and prefer creating raw volumes_. Minimum volume sizing is as below:
+
+Cluster size | Minimum Size of Volume | Minimum number of volumes|
```
bob-bins: Somewhat of a nitpick, but I'd suggest:
```
Volume requirements for a default setup are shown below:

Cluster size | Minimum volume size | Minimum number of volumes |
```

configurator/README.md
```diff
@@ -185,3 +185,50 @@ https://raw.githubusercontent.com/draios/sysdigcloud-kubernetes/Templating_k8s_c
 - Modify the values.yaml
 - Copy the tar file to the directory
 - Run the tar file `bash sysdig_configurator.tar.gz`
+
+### Miscellaneous
```
bob-bins: I agree with you that Miscellaneous is an odd title for this section. Why not name it `Local Storage`?

PR #247, opened 2019-07-10
bump backend version

bob-bins:
2304 is the version from the previous release so it does not have the secure Elasticsearch auth code


papajulio:
This PR it's not needed here for this branch, we will be testing different candidate versions that we are building, no problem in having the latest stable there until we have the new stable and we generate the PR. Our deployment scripts replace that so no issue in having latest stable here. Thanks for the fix anyway

janoe:
Maybe this version update makes more sense in this files related to the configurator:
https://github.com/draios/sysdigcloud-kubernetes/commit/bf73865995fb8111d2de80a03b9c639d8a70dbd2

PR #257, opened 2019-07-12
Devops 1804 template java certs

jpravin:
Adding support for custom-ca.


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -122,8 +122,20 @@ if [[ "$DNS_NAME" != "$COMMON_NAME" ]]; then
...
   fi
 fi
+
 set -e #re-enable exit on error
 
+CUSTOM_CA=$(yq -r .sysdig.customCa "$TEMPLATE_DIR/values.yaml")
+if [[ $CUSTOM_CA == "true" ]]; then
+  CUSTOM_CERT=$MANIFESTS/certs/custom-ca.pem
+  if [[ ! -f $CUSTOM_CERT ]]; then
+    log error "Custom certificate is set but not provided. Please copy the custom certificate to {MANIFESTS}/certs/custom-ca.pem"
```
bob-bins: should use `Please copy the custom certificate to ${CUSTOM_CERT}` instead of `Please copy the custom certificate to {MANIFESTS}/certs/custom-ca.pem`

related question - from the container's perspective it's `$CUSTOM_CERT`, but from the user's perspective wouldn't it be a path from the dir that they mounted? If so this error message might be confusing to them.

configurator/sysdig-chart/tests/resources/large_aws/values.yaml
```diff
@@ -82,5 +82,9 @@ sysdig:
   #Override for default collector port
   #defaults to 6443
   collectorPort:
+  #custom ca(certificate authorities)
+  #adds custom ca's to trusted list
```
bob-bins: the apostrophe is not grammatically correct here :P
```
# Adds custom certificate authorities to trusted list
# If true, expected certs/custom-ca.pem in the mounted manifests directory
```

configurator/sysdig-chart/tests/resources/large_aws_monitor_only/values.yaml
```diff
@@ -82,5 +82,9 @@ sysdig:
   #Override for default collector port
   #defaults to 6443
   collectorPort:
+  #custom ca(certificate authorities)
+  #adds custom ca's to trusted list
```
bob-bins: same comment as above

configurator/sysdig-chart/generate_templates.sh
jpravin: Mike i am working on another pr. I will rebase this on that and fix this.

configurator/sysdig-chart/tests/resources/large_aws/values.yaml
jpravin: reabse will remove this config.

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -100,6 +105,10 @@ else
   fi
 fi
 
+if [[ -d "$MANIFESTS/java-certs" ]]; then
+  cp -r "$MANIFESTS/java-certs" "$MANIFESTS_TEMPLATE_BASE/common-config/"
```
bob-bins: line 108 is identical to 65. Can we do both these copies in a single `if` block?

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -61,6 +61,11 @@ if [[ -z "$(ls -A "${MANIFESTS}/elasticsearch-tls-certs")" ]]; then
     mv out "${MANIFESTS}/elasticsearch-tls-certs")
 fi
 
+#JavaCerts exists in manifests dir copy them to helm templates
+if [[ -d "$MANIFESTS/java-certs" ]]; then
+  cp -r "$MANIFESTS/java-certs" "/sysdig-chart/"
```
bob-bins: we can use `"$TEMPLATE_DIR"` instead of `"/sysdig-chart/"`

jpravin: I dont this this seqment is needed. Let me verify this.

configurator/sysdig-chart/values.yaml
```diff
@@ -79,5 +79,9 @@ sysdig:
   #Override for default collector port
   #defaults to 6443
   collectorPort:
+  #custom ca(certificate authorities)
+  #adds custom ca to trusted list
+  #true expects certs/custom-ca.pem in manifest directory
```
bob-bins: What do you think about
```
# If true, expects certs/custom-ca.pem in the current working directory, which will be added to the trust store
```
instead?

PR #259, opened 2019-07-16
Introduce image release workflow.

bjhaid:
## What does this do

- On every successful(tests passing) build of the main brainch(Templating_k8s_configurations):
  - pushes a git tag matching next_version-rc{JENKINS_BUILD_NUMBER}
  - builds a docker and tags a docker image matching the git tag and pushes to the internal registry
- Introduces full release workflow that is depicted in the README.
- Cleanup Jenkinsfile to use functions in all places where duplicate variables have been used.

## Note for Reviewers
- Please read through the workflow in the README and suggest
improvements.


wpalmeri:
shouldn't the non-rc tag (and push) be invoked manually rather than an automatic weekly job? i.e. we should feel confident in a vetted build to cut a new release tag

bjhaid:
> shouldn't the non-rc tag (and push) be invoked manually rather than an automatic weekly job? i.e. we should feel confident in a vetted build to cut a new release tag

There's nothing automating the release, it's all manual the PR body is just documenting the process, I should maybe update it to reflect that :)

Jenkinsfile
```diff
@@ -46,13 +46,19 @@ pipeline {
   }
   stages {
     stage('ShellCheck') {
+      when {
+        not { tag pattern: "^v\\d+\\.\\d+\\.\\d+\$", comparator: "REGEXP" }
```
bob-bins: can we turn these regexes into variables with descriptive names? E.g. `semver`, `semverWithPrerelease`, `semverWithPrereleaseAndBuild`, etc.

configurator/README.md
```diff
@@ -45,6 +45,75 @@ The rationale for the workflow is it helps the code reviewers to see the
 effect(s) of the changes introduced, it also helps exercise the config
 generation workflow.
 
+## Releasing
+
+### Release Candidates
+
+Every workday morning (Pacific Time), the release personnel tags a release
```
bob-bins: Why every workday morning, instead of "as appropriate"?

bjhaid: We should be creating rcs as soon as we have at least one commit but also not so frequently that it becomes painful... every day seems like a convenient interval, but then it's not cast in stone and folks can always be flexible when and where needed.

In the end I want to make this a Jenkins job and the interval and pain becomes less relevant

configurator/README.md
```diff
@@ -45,6 +45,75 @@ The rationale for the workflow is it helps the code reviewers to see the
...
+candidate with the below:
+
+- Run the below command
+
+```bash
+CURRENT_CONFIGURATOR_RC_VERSION=$(cat configurator/rc_version)
+if [[ ! -z $(git log $CURRENT_CONFIGURATOR_RC_VERSION..) ]]; then
+  RC_VERSION=${CURRENT_CONFIGURATOR_RC_VERSION#*rc}
+  NEXT_VERSION=${CURRENT_CONFIGURATOR_RC_VERSION%-rc*}
```
bob-bins: if `CURRENT_CONFIGURATOR_RC_VERSION=0.0.1-rc0` then `NEXT_VERSION=0.0.1`. Would `CURRENT_VERSION` be a better variable name?

configurator/README.md
```diff
@@ -45,6 +45,75 @@ The rationale for the workflow is it helps the code reviewers to see the
...
+  NEXT_RC_VERSION=$((1 + $RC_VERSION))
+  NEXT_CONFIGURATOR_RC_VERSION="$NEXT_VERSION-$RC_VERSION"
```
bob-bins: do you mean `NEXT_CONFIGURATOR_RC_VERSION="$NEXT_VERSION-$NEXT_RC_VERSION"` here?

bjhaid: ah yeah thanks

configurator/README.md
bjhaid: no, current released version is 0.0.0 and next release is 0.0.1

configurator/README.md
bob-bins: I may not have full context here, so to confirm my understanding is this intended to be run only from the `Templating_k8s_configurations` branch? When you say `In the end I want to make this a Jenkins job and the interval and pain becomes less relevant` do you mean in the future this would be a Jenkins job that runs automatically on a commit to `Templating_k8s_configurations`?

bjhaid: > so to confirm my understanding is this intended to be run only from the Templating_k8s_configurations branch

Yes Templating_k8s_configurations today or whatever the main branch for configurator work is in the future

> do you mean in the future this would be a Jenkins job that runs automatically on a commit to Templating_k8s_configurations?

I meant a jenkins job that runs on some schedule e.g every day

bob-bins: Wondering why doing this on a regular schedule is desired rather than be event-based like on a commit/merge. There might not be a change every day so we'd end up pushing rcs with no changes vs the previous rc if it's on a regular interval.

bjhaid: > There might not be a change every day so we'd end up pushing rcs with no changes vs the previous rc if it's on a regular interval

It will be a noop.

> And IMO it would be better to have a rc for every PR rather than one for multiple on days where more than 1 PR gets merged.

This is fair, maybe we don't even need rc tag altogether and maybe just push an rc image on every successful build of the main branch

Jenkinsfile
mbreitung: are we putting a `v` in front of the version?

configurator/README.md
bob-bins: I fully agree about having a way to link a values.yaml to a particular configurator commit/image. And I agree that an rc tag is a fine way of doing that.

configurator/next_version
```diff
@@ -0,0 +1 @@
+v0.0.1
```
bob-bins: I think you accidentally left in the `v`

bjhaid:
This PR is ready for review, folks should be please re-review. Thanks!

bjhaid:
> Is there an action to include this new work in the docs? Minus that question /lgtm

README updated here:

https://github.com/draios/sysdigcloud-kubernetes/pull/259/files#diff-ac198bcdc528181106e9daf3a6a09fc9

configurator/README.md
```diff
@@ -45,6 +45,64 @@ The rationale for the workflow is it helps the code reviewers to see the
 effect(s) of the changes introduced, it also helps exercise the config
 generation workflow.
```
bob-bins: It would be nice to have a section that explains the format of the version, once it's been finalized. Something like:

```
## Version

The Configurator version follows the format:

	<sysdig_platform_version>-<configurator_semver_version>

`uber` images contain all assets necessary to run the Configurator for air-gapped installations, and will have `-uber` appended to the end of the version.
```

configurator/README.md
```diff
@@ -68,6 +70,63 @@ installations](#full-airgap-installation) the versioning scheme is:
...
+indicate the latest release tag.
+- Update the internal wiki instructions pointing at the latest release.
+- Update `configurator/current_version` to reflect the new tag, e.g:
+```
+cat configurator/next_version > configurator/current_version
+```
+- Reset the `configurator/next_version` file to point at some hypothetical
+future version that has not been previously used, e.g:
+```bash
+echo -n 2.3.0-0.1.1 > configurator/rc_version
```
bob-bins:
```suggestion
echo -n 2.3.0-1 > configurator/rc_version
```

configurator/current_version
```diff
@@ -0,0 +1 @@
+2.3.0-0.0.0
```
bob-bins:
```suggestion
2.3.0-1
```

configurator/next_version
```diff
@@ -0,0 +1 @@
+2.3.0-0.0.1
```
bob-bins:
```suggestion
2.3.0-2
```

Jenkinsfile
```diff
@@ -74,12 +108,15 @@ pipeline {
...
+      when {
+        not { equals expected: nextReleaseTag(), actual: env.TAG_NAME }
+      }
       steps{
         script {
           docker.withRegistry("https://quay.io", "QUAY") {
             sh(
               "cd configurator && " +
-              "make test_uber_tar"
+              "IMAGE_NAME=${dockerImage()} make test_uber_tar"
```
bob-bins: should this be `uberImage` instead of `dockerImage`?

configurator/README.md
bob-bins: resolved since this was added in another PR

Jenkinsfile
bjhaid: no, it's the name of the configurator image, see: 
https://github.com/draios/sysdigcloud-kubernetes/blob/c275c8f83fbf7a36268c19424886557931c8eb4b/configurator/Makefile#L1

for tests this can actually be removed... the name of the image really does not matter for the test and the Makefile provides a sane default, it's only useful for pushing and in that case it needs to be combined with `UBER_IMAGE_NAME` which the push piece does.

configurator/README.md
bjhaid: actually `rc_version` is no longer a thing, I need to read through the versioning/release portion of this README again and correctly update it, I merged upstream to make it resolve conflicts, I should not have accepted the change

bob-bins: ah ok. Lmk when you're done fixing up the README and I'll take another look! I still had the old model in mind

configurator/README.md
```diff
@@ -68,6 +70,64 @@ installations](#full-airgap-installation) the versioning scheme is:
...
+```
+- Create a new release tag, e.g:
+```bash
+git tag -F <(git log --oneline $(cat configurator/current_version)..) $(cat configurator/next_version)
+git push origin refs/tags/"$(cat configurator/next_version)"
+```
+- Go the Jenkins UI and click the build now button for the new tag (this will
+not be necessary when we move to the new Jenkins).
+- Wait for the Jenkins build for the new tag to complete successfully.
+- Update every part of the README.md(this file) that indicates a version to
```
bob-bins: IMO this step isn't completely necessary, since people shouldn't be looking at this readme to see what the latest version is (they can just look at the releases tab once this is moved to its own repo)

configurator/README.md
```diff
@@ -68,6 +70,64 @@ installations](#full-airgap-installation) the versioning scheme is:
...
+indicate the latest release tag.
+- Update the internal wiki instructions pointing at the latest release.
+- Update `configurator/current_version` to reflect the new tag, e.g:
+```
+cat configurator/next_version > configurator/current_version
+```
+- Increment the number after the hypen e.g `1` in `2.3.0-1` by `1` and update
```
bob-bins: This is true if the platform version does not change, but it may be good to mention the case when it does, e.g. `2.3.0-3` to `2.3.1-1`

configurator/README.md
```diff
@@ -68,6 +70,64 @@ installations](#full-airgap-installation) the versioning scheme is:
...
+On every successful build of the main branch, Jenkins tags a release candidate
+version matching `<next_release_tag>-rc<$JENKINS_BUILD_NUMBER>` and pushes
+the git tag, and a docker image matching the git tag.
+
+### Full Release
+
+The workflow for doing a full (non-rc) release is as below:
+
+- Change `configurator/next_version` to the new tag to be released, for
+example if the new tag is `2.3.0-1`, do as below:
```
bob-bins: this is technically true, but a bit of an odd example since `2.3.0-1` is our first release :P

bob-bins:
looks pretty good to me! just have 2 commit suggestions further above and a couple non-blocking comments. other than that I'd be ready to approve

configurator/README.md
bjhaid: It still is necessary, the docker image should match the README version

bob-bins: yeah you're right, in the Usage section. My bad!

PR #260, opened 2019-07-16
Devops 1987 test override framework

jpravin:
This uses inheritance model for values.yaml. 

The values.yaml in sysdigcloud is used as a base. And tests can provide an override values.yaml.


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -160,15 +177,15 @@ else
   log info "skipping step7d: data-stores postgres - needed only for secure"
 fi
 
-IS_REDIS_HA=$(yq .sysdig.redisHa "$TEMPLATE_DIR/values.yaml")
-if [[ ${IS_REDIS_HA} == "false" ]]; then
```
bjhaid: This change seems to have introduced a significant amount of redis related diff, can you confirm the change is as expected?

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -3,14 +3,26 @@
 # Allow files generated in container be modifiable by host user
 umask 000
 
+#defaults to no overrides
+VALUES_OVERRIDE=""
+while getopts ":f:" opt; do
+  case $opt in
+    f) VALUES_OVERRIDE="$OPTARG"
+    ;;
+    \?) echo "Invalid option -$OPTARG" >&2
```
bob-bins: It is impossible to google what `\?` means. Can you state it here please?

configurator/sysdig-chart/shared-values.sh
```diff
@@ -2,7 +2,22 @@
 
 export TEMPLATE_DIR="/sysdig-chart"
 
-K8S_NAMESPACE="$(yq -r .namespace $TEMPLATE_DIR/values.yaml)"
+function readYaml() {
+  local valueToRead=$1
+  local valueOverride=""
+
+  if [[ $# -eq 2 ]]; then
```
bob-bins: no need for this block - `valueOverride=$2` will automatically default to `''` if a 2nd parameter is not passed in

configurator/sysdig-chart/shared-values.sh
```diff
@@ -2,7 +2,22 @@
 
 export TEMPLATE_DIR="/sysdig-chart"
 
-K8S_NAMESPACE="$(yq -r .namespace $TEMPLATE_DIR/values.yaml)"
+function readYaml() {
```
bob-bins: I think we can improve some of the naming around this function. My example might not be ideal, but should illustrate the intent.
```
function valueFromValuesYaml() {
  local keyPath=$1
  local pathToValuesOverride=$2

  if [[ "$pathToValuesOverride" == "" ]]; then
    yq -r "$keyPath" "$TEMPLATE_DIR/values.yaml"
  else
    yq -r -s ".[0] * .[1] | $keyPath" "$TEMPLATE_DIR/values.yaml" "$pathToValuesOverride"
  fi
}
```

configurator/sysdig-chart/generate_templates.sh
jpravin: Yes. Ayo This is expected because if the value is not set it defaults RedisHa to true. I changed the defaults from true to false.

configurator/sysdig-chart/generate_templates.sh
jpravin: Any argument  which is not caught above.

bob-bins: so is it identical to `*` here?

configurator/sysdig-chart/shared-values.sh
jpravin: If i dont do that i ran into this,
`  
  local valueOverride=$2

  #if [[ $# -eq 2 ]]; then
  #  valueOverride=$2
  #fi
`
/sysdig-chart/shared-values.sh: line 7: $2: unbound variable

configurator/sysdig-chart/shared-values.sh
jpravin: i will make it readConfigFromValuesYaml.

configurator/sysdig-chart/shared-values.sh
bob-bins: ah i see. should bypass that error `valueOverride=${2:-''}`

configurator/sysdig-chart/generate_templates.sh
jpravin: yes more like any

configurator/sysdig-chart/shared-values.sh
jpravin: cool i can add that. That makes it simple

PR #261, opened 2019-07-17
Add support for nodeaffinity.

bjhaid:
It introduces a new configuration option `nodeaffinityLabel`, can be
configured as below:

```yaml
nodeaffinityLabel:
  key: 'sysdigcloud.node'
  value: true
```

This reflects a key and value represents keys and values of node labels that
should have been previously applied to nodes intended to run sysdig pods.

The nodeaffinityLabel gets applied as a [node
affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity)
to all sysdigcloud pods, influencing k8s decision to schedule them to
only nodes with the label.


bjhaid:
> Everything looks good. Would it be good to put a blurb in the readme regarding this?

There's some upcoming work to document everything in the values.yaml file in the README, so I'm deferring adding a README blurb to that work.

jpravin:
Ayo my only question is should we make this configurable.

configurator/sysdig-chart/templates/data-stores/cassandra/cassandra-statefulset.yaml
```diff
@@ -25,6 +25,10 @@ spec:
         app: sysdigcloud
         role: cassandra
     spec:
+      {{ if .Values.useNodeSelector | default false }}
+      nodeSelector:
+        sysdigcloud.node: true
```
mbreitung: shouldn't the name `sysdigcloud.node` be templated? Customers will name this whatever they want

jpravin: At what granularity do we need this? Should each app have a unique selector?

wpalmeri:
from the k8s docs below, should we use nodeAffinity instead?

> nodeSelector continues to work as usual, but will eventually be deprecated, as node affinity can express everything that nodeSelector can express.

jpravin:
> from the k8s docs below, should we use nodeAffinity instead?
> 
> > nodeSelector continues to work as usual, but will eventually be deprecated, as node affinity can express everything that nodeSelector can express.

But are we planning to provide a configurable array of labels for every app ?

bjhaid:
> We need to make the nodeSelector key:value pair configurable. I talked with Jan and he did confirm that we can assume they will take care of ensuring nodes are labelled.

Addressed

> from the k8s docs below, should we use nodeAffinity instead?

Done

> But are we planning to provide a configurable set of labels for every app ?

I do not think we should do this, this means we are doing scheduling of pods by hand instead of relying on k8s to figure out how best to schedule pods

PR #265, opened 2019-07-19
DEVOPS-2004-Adding certs configurations

jpravin:


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -141,15 +137,17 @@ if [[ "$DNS_NAME" != "$COMMON_NAME" ]]; then
 fi
 set -e #re-enable exit on error
 
-CUSTOM_CA=$(yq -r .sysdig.customCa "$TEMPLATE_DIR/values.yaml")
+CUSTOM_CA=$(yq -r .sysdig.certificate.customCa "$VALUES_OVERRIDE")
```
bob-bins: should we use `readConfigFromValuesYaml` instead?

configurator/sysdig-chart/test.sh
```diff
@@ -42,9 +42,24 @@ function concat_config() {
...
   for directory in /sysdig-chart/tests/resources/*/; do
+    run_test "$directory"
+  done
+}
+
+function clear_and_copy_manifests(){
+  local directory=$1
+  rm -rf /manifests/*
+  if [[ ! -d /manifests ]]; then
+    mkdir /manifests
```
bob-bins: can just do `mkdir -p /manifests` instead of the `if` block

bob-bins: actually maybe we can just
```
rm -rf "$MANIFESTS"
mkdir "$MANIFESTS"
cp -r "$directory"/* "$MANIFESTS"
```

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -70,32 +70,28 @@ fi
...
-    cp "$KEY_FILE" "$MANIFESTS_TEMPLATE_BASE/common-config/certs/server.key"
+  log info "Certificates are not present."
+  if [[ "$GENERATE_CERTIFICATE" = "true" ]]; then
+    log info "Generating new certificate"
+    openssl req -new -newkey rsa:2048 -days 3650 -nodes -x509 -subj "/C=US/ST=CA/L=SanFrancisco/O=ICT/CN=$DNS_NAME" -keyout "$KEY_FILE" -out "$CERT_FILE"
+    log info "Certificates are generated. Copying the generated certs"
+    cp "$CERT_FILE" "$KEY_FILE" "$MANIFESTS_TEMPLATE_BASE/common-config/certs/"
   else
-    log error "Cannot find certificate files. Exiting"
+    log error "Cannot find certificates and Certification generation is set to false. Exiting"
```
bjhaid: s/Certification/Certificate

configurator/sysdig-chart/test.sh
bjhaid: you can't always do `rm -rf "$MANIFESTS"`, because it could be a bind-mounted directory, the only time when it is not bind-mounted is in test or config_gen, but I agree `mkdir -p` should be fine

configurator/sysdig-chart/generate_templates.sh
jpravin: Damn.. I started working on this before the other PR and completely forgot that change.
Thanks Mike.

configurator/sysdig-chart/test.sh
jpravin: Done 👍

configurator/sysdig-chart/generate_templates.sh
jpravin: fixed 👍

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -70,32 +70,28 @@ fi
...
-GENERATED_KEY=$MANIFESTS/certs/server.key
+CERT_FILE=$MANIFESTS/certs/server.crt
+KEY_FILE=$MANIFESTS/certs/server.key
 DNS_NAME=$(readConfigFromValuesYaml .sysdig.dnsName "$VALUES_OVERRIDE")
 
 mkdir "$MANIFESTS_TEMPLATE_BASE/common-config/certs"
-if [ ! -d "$MANIFESTS/certs" ]; then
-  log info "Making certs manifests dir"
+if [[ ! -d "$MANIFESTS/certs" ]]; then
+  log info "Creating certs manifests dir"
```
bob-bins:
```suggestion
  log info "Creating $MANIFESTS/certs dir"
```

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -141,15 +137,17 @@ if [[ "$DNS_NAME" != "$COMMON_NAME" ]]; then
...
-  if [[ ! -f $CUSTOM_CERT ]]; then
+  CUSTOM_CERT="$MANIFESTS"/certs/custom-ca.pem
+  if [[ ! -f "$CUSTOM_CERT" ]]; then
     log error "Custom ca is set but not provided. Please provide a custom ca cert at certs/custom-ca.pem in the current working directory."
   else
     log info "Copying custom ca to $MANIFESTS_TEMPLATE_BASE/common-config/certs/"
     cp "$CUSTOM_CERT" "$MANIFESTS_TEMPLATE_BASE/common-config/certs/"
   fi
+else
+  log info "Custom Ca not defined. Moving on. $CUSTOM_CA"
```
bob-bins: IMO this `else` block isn't necessary. If custom_ca == true, the `if` block will be guaranteed to print something out that indicates a custom CA is desired. The lack of that implies custom_ca == false

But if you prefer having the else block, I'd suggest `log info "Custom CA is set to false. Continuing..."` Since "Custom Ca not defined" might be interpreted as "Custom Ca should be defined, but isn't"

configurator/sysdig-chart/test.sh
```diff
@@ -42,9 +42,22 @@ function concat_config() {
 
 function run_tests() {
   for directory in /sysdig-chart/tests/resources/*/; do
+    run_test "$directory"
+  done
+}
+
+function clear_and_copy_manifests(){
+  local directory=$1
+  rm -rf /manifests/*
```
bob-bins: may be good to use `$MANIFESTS` here instead of `/manifests`

configurator/sysdig-chart/values.yaml
```diff
@@ -50,9 +50,12 @@ sysdig:
   #sure the key and cert are in your current working
   #directory
   certificate:
+    #false expects certs at certs/server.key and certs/server.crt in current working directory
     generate: true
```
bob-bins: hmm if this is `true` and `customCa` is also true, will it generate certificates and just not use them?

jpravin: customCa and cert generation are 2 different features.
customCa is for java process to communicate with LDAP servers, cassandra and es ssl.
Certificate generate:true is for ssl cert presented by ui.

configurator/sysdig-chart/test.sh
jpravin: This file does not import shared-values.sh. I can add that reference.

bjhaid:
I am going to close this, the work in https://github.com/draios/sysdigcloud-kubernetes/pull/310 demonstrates this convention might not be right course of action.

PR #267, opened 2019-07-22
Put back elasticsearch url to config.

janoe:
In order to avoid errors when, for example, customer don't upgrade elasticsearch, we can let elasticsearch.url entry in configmap for the moment, to be safer, and we can remove it in next releases if we want to.

This is related too with this change in the docs:
https://sysdig.atlassian.net/browse/DOC-764

@bob-bins , does this makes sense to you?


sysdigcloud/config.yaml
```diff
@@ -104,6 +104,11 @@ data:
   # Required: Elasticsearch URL. If Elasticsearch is deployed as a Kubernetes service,
   # this will be the service name. If using an external database, put the proper address
   elasticsearch.hostname: sysdigcloud-elasticsearch
+  # Put back this entry to be backward compatible with ES
+  # with no auth plugin activated.
+  # This entry can be removed if we use our ES version:
```
mbreitung: s/can/should

janoe: Replaced

sysdigcloud/config.yaml
```diff
@@ -104,6 +104,11 @@ data:
   # Required: Elasticsearch URL. If Elasticsearch is deployed as a Kubernetes service,
   # this will be the service name. If using an external database, put the proper address
```
bob-bins: `If using an external database, put the proper address` should be removed - our code adds `http://` or `https://` to construct the full URL based on the value of `elasticsearch.searchguard.enabled`.

sysdigcloud/config.yaml
```diff
@@ -104,6 +104,11 @@ data:
...
   elasticsearch.hostname: sysdigcloud-elasticsearch
+  # Put back this entry to be backward compatible with ES
+  # with no auth plugin activated.
```
bob-bins: IMO lines 107 and 108 aren't necessary. We can be a bit more concise if we say `This entry is for backward-compatibility and should be removed when using backend version >= {{RELEASE_VERSION}}`

Regarding the change from `ES version` to `backend version`, it's because the backend init script is the component that uses this field. The ES container does not.

janoe: Yes, it was my misunderstanding, I thought that ES container needed this parameter by my conversations with customer success, but I've double checked this and it's not necessary.

PR #277, opened 2019-07-26
Introduce sane defaults for most of values in values.yaml

bjhaid:
This introduces a bunch of sane defaults for fields in values.yaml that previously did not have defaults, see https://github.com/draios/sysdigcloud-kubernetes/pull/277/files#diff-f5fca167e6f0e46fe6fcfb827cbd74e8 for adequate coverage of the defaults

It also makes `readConfigFromValuesYaml` consider defaultValues.yaml when looking up values and makes the logic for cassandra multi-az ec2snitch smart enough to handle when cloudProvider is not configured


configurator/sysdig-chart/defaultValues.yaml
```diff
@@ -1,3 +1,8 @@
+#defines which scripts needs to be run
+#generate - performs templating & kustomization
+#deploy   - apply's the generated script in k8s environment
+#generate+deploy - combination of the previous 2 commands.
+scripts: generate+deploy
```
bob-bins: would it be better to use yaml arrays rather than create our own string parsing convention?
```
scripts:
  - generate
  - deploy
```

bjhaid: I considered this... while the code that consumes it today uses a list approach to consume it, it technically is an ENUM and not a list, I am already considering adding options such as `upgrade` and `delete`... I think eventually we should have `generate`, `create`, `update` and `delete` as options with `create`, `update`, and `delete` each a superset of `generate`

bjhaid: I reverted this to what it was before "generate deploy", we can in a subsequent PR/workstream change this appropriately

bjhaid:
This PR has changed somewhat since it was opened so I'll like if folks can re-review.

This now defaults a bunch of things, that previously did not have a default. I am going to update the PR description to reflect this

configurator/sysdig-chart/defaultValues.yaml
```diff
@@ -1,10 +1,25 @@
+apps: monitor secure
+scripts: generate deploy
+namespace: sysdigcloud
```
bob-bins: can we change this to `sysdig` instead of `sysdigcloud`?

bjhaid: oh I intend doing this, but in a separate PR, because it is going to introduce a bunch of diffs in the test code and I don't want to confuse folks reviewing the PR

PR #284, opened 2019-07-30
Don't show dir not found error to user

marojor:
Redirect stderr from elasticsearch cert dir check to `/dev/null`

This silences this noise in the configurator run:

```
step3.5: creating elasticsearch certs for Searchguard
ls: cannot access '/manifests/elasticsearch-tls-certs': No such file or
directory
Generating certs for Searchguard...
```


configurator/sysdig-chart/generate_templates.sh
```diff
@@ -54,7 +54,7 @@ else
 fi
 
 log info "step3.5: creating elasticsearch certs for Searchguard"
-if [[ -z "$(ls -A "${MANIFESTS}/elasticsearch-tls-certs")" ]]; then
+if [[ ! -d "${MANIFESTS}/elasticsearch-tls-certs" ]]; then
```
bjhaid: this actually should be:

 `if [[ -z "$(ls -A "${MANIFESTS}/elasticsearch-tls-certs" 2> /dev/null)" ]]; then`

because we want to test that the directory exist and is not empty, and we can silence the noise by redirecting stderr to `/dev/null`, this probably needs a comment so someone else does not make the same change in the future

bjhaid: @marojor do you want to fix this, or you want me to fix?

marojor: @bjhaid I wasn't sure whether to fix it here or close this PR and open a new (clean) one and in the meantime got hit by something else.

It would be great if you can fix it.

bjhaid: Done in:

https://github.com/draios/sysdigcloud-kubernetes/pull/284/commits/fc91a4fa56d2ab1eb09d65514e3753140557e9b1

PR #285, opened 2019-07-30
Change default namespace to 'sysdig'

bjhaid:
This also makes sure the test code uses most of the new defaults.


bjhaid:
ah I did not mean to close this, I was going to merge, dangit

PR #287, opened 2019-07-30
 Document values.yaml fields.

bjhaid:
This documents all of the possible configuration of values.yaml with
their descriptions, options and default values, all options that are
required come early in the documentation while the rest follow.


bjhaid:
https://github.com/draios/sysdigcloud-kubernetes/pull/287/files?short_path=b9f38ad#diff-b9f38ad3b80210edeb528a661786881c

is a good link to review the configuration.md as it will render

configurator/configuration.md
```diff
@@ -0,0 +1,56 @@
...
+sysdig.elasticsearchVersion | Version of ElasticSearch run by sysdig cloud | - | `5.6.44.0` |
+sysdig.postgresVersion | Version of PostgreSQL run by sysdig cloud | - | `10.6.10` |
+sysdig.redisVersion | Version of Redis run by sysdig cloud | - | `4.0.12.5` |
+sysdig.redisHaVersion | Version of HA Redis run by sysdig cloud | - | `4.0.12.5-ha` |
+sysdig.haproxyVersion | Version of HAProxy run by sysdig cloud | - | `v0.7-beta.7` |
+sysdig.rsyslogVersion | Version of rsyslog run by sysdig cloud | - | `8.34.0.5` |
+sysdig.localVolumeProvisioner | Version of localVolumeProvisioner run by sysdig cloud | - | `v2.3.2` |
+elasticsearch.searchguard.enabled | Enables user authentication and TLS-encrypted data-in-transit with [Searchguard](https://search-guard.com/) | `true \| false` | `true` |
+elasticsearch.searchguard.adminUser | The user bound to the ElasticSearch Searchguard admin role | - | `sysdig` |
+elasticsearch.searchguard.verifySSL | Enables hostname verification when the backend components communicate with Elasticsearch nodes | `true\|false` | `false` |
```
bob-bins:
```suggestion
elasticsearch.searchguard.verifySSL | Enables hostname verification when the backend components communicate with Elasticsearch nodes. Only applicable if elasticsearch.searchguard.enabled is `true` | `true\|false` | `false` |
```

Though on second thought, this option must be false unless the customer is allowed to add their own ES ca-cert, which they don't yet have the option to for the configurator. So perhaps it shouldn't be a configurable parameter.

configurator/configuration.md
```diff
@@ -0,0 +1,56 @@
...
+quaypullsecret | quaypullsecret provided by the marketing team that kubelet will use to pull sysdigcloud images from Quay | - | - |
+storageClassProvisioner | name of [storage class provisioner](https://kubernetes.io/docs/concepts/storage/storage-classes/#provisioner) to use when creating the configured `storageClassName` parameter. hostPath is only supported in cluster configured with `size` parameter of `small`, it should only be used for experimentation and not in production. `local` should be used in clusters that do not have a provisioner, see [local storage](README.md#local-storage) for instructions on setting it up | `aws\|gke\|hostPath\|local` | - |
+sysdig.agentCount | number of sysdig agents to deploy | - | - |
+sysdig.license | sysdig license as provided by the marketing team | - | - |
+sysdig.dnsName | Domain name the sysdig api will be served on| - | - |
+apps | defines applications needed. Monitor installed by default.<br> Secure is optional. Remove secure if it is not deployed> |`monitor\|monitor secure` | `monitor secure` |
```
bob-bins: should this have `monitor` be the default option, given that not everyone is using `secure`? And I would remove the redundant `Secure is optional. ` since it's already stated in `Options`

configurator/configuration.md
```diff
@@ -0,0 +1,56 @@
...
+sysdig.cassandraReplicaCount | Number of cassandra replicas, this is a noop for small clusters| - | small cluster: 1<br>medium cluster: 3<br>large cluster: 6|
+sysdig.elasticSearchReplicaCount | Number of elasticsearch replicas, this is a noop for small clusters| - | small cluster: 1<br>medium cluster: 3<br>large cluster: 6|
+sysdig.apiReplicaCount | Number of sysdig api replicas, this is a noop for small clusters| - | small cluster: 1<br>medium cluster: 3<br>large cluster: 5|
+sysdig.collectorReplicaCount | Number of sysdig collector replicas, this is a noop for small clusters| - | small cluster: 1<br>medium cluster: 3<br>large cluster: 5|
+sysdig.workerReplicaCount | Number of sysdig worker replicas, this is a noop for small clusters| - | small cluster: 1<br>medium cluster: 3<br>large cluster: 5|
+sysdig.redisHa | determines if redis should run in HA mode| `true\|false` | `false`|
+sysdig.smtpServer | specifies smtp server to use to send email| - | - |
+sysdig.smtpServerPort | specifies port for the configured smtp server| - | `25` |
```
bob-bins: should we specify some port range, like we do for some of the other port parameters?
```suggestion
sysdig.smtpServerPort | specifies port for the configured smtp server| `1024-65535` | `25` |
```

configurator/configuration.md
```diff
@@ -0,0 +1,56 @@
...
+sysdig.smtpUser | specifies user for the configured smtp server| - | - |
+sysdig.smtpPassword | specifies password for the configured smtp user| - | - |
+sysdig.smtpProtocolTLS | specifies if TLS should be used for smtp| `true\|false` | - |
+sysdig.smtpProtocolSSL | specifies if SSL should be used for smtp| `true\|false` | - |
+sysdig.smtpFromAddress | specifies the email address for the FROM field of sent emails| - | - |
+sysdig.openshiftUrl | specifies openshift http url with port number  | - | - |
```
bob-bins: should we mention these are required(?) if and only if `deployment` is `openshift`?

configurator/configuration.md
bjhaid: I can just delete it and not document it, what do you think?

configurator/configuration.md
jpravin: SMTP can be run on any port. It need not be a high port.

bjhaid: thanks! I update. This is technically `1-65535`, since the canonical port for non-TLS smtp is 25, then 465 and 587 for TLS smtp.

configurator/configuration.md
bob-bins: I'm fine with deleting it from the docs! (but still leaving as a parameter in the implementation)

configurator/configuration.md
bob-bins: openshiftUsername and openshiftPassword can have the same addition as well

bjhaid: done

configurator/configuration.md
bjhaid: > should this have monitor be the default option, given that not everyone is using secure? 

Kini advised that `monitor+secure` should be the default

PR #289, opened 2019-07-31
Specify mounted values.yaml as override for SCRIPTS.

bjhaid:
When scripts was defaulted in
https://github.com/draios/sysdigcloud-kubernetes/pull/277, it did not
specify the override for the `scripts` parameter causing cases where
values.yml does not defined `scripts` to break. This addresses this
problem by making `readConfigFromValuesYaml` consider the
defaultValues.yaml along with the mounted `values.yaml`

See
https://github.com/draios/sysdigcloud-kubernetes/blob/46677c9d89694515c7f7b3c88d44ba68e3464283/configurator/sysdig-chart/shared-values.sh#L9-L13
for context on how and why all of these works.


bjhaid:
> Isn't the same problem happening in generate_templates.sh?

Great catch! I have addressed that also in:

https://github.com/draios/sysdigcloud-kubernetes/pull/289/commits/3e6ef6b244ae7be811367f3fee3e214e058ad545

bjhaid:
> Now that we are using the override option for even the regular installs. Can we remove the -f option and make it a mandatory argument?

Yes

> We can even remove the override check in shared-values.sh.

Also yes.

I'll do both and update this PR shortly

bjhaid:
> I'll do both and update this PR shortly

Done in: https://github.com/draios/sysdigcloud-kubernetes/pull/289/commits/6b059e5630a945c904158019b374cec25d535405

PR #293, opened 2019-08-02
formatting change and pulling in changes from onprem-2019-07

bob-bins:


bob-bins:
closing this and creating one that branches off of `dev`

PR #302, opened 2019-08-06
change steps for creating searchguard users to applying a yaml file i…

bob-bins:
…nstead of running a kubectl create secret cmd

Some customers prefer having as much of the sysdig platform in k8s yamls as possible, since it makes it easier if they must recreate their environment.


bob-bins:
@bjhaid it's a perfectly valid point. The value to the secret needs to be a valid base64 encoding, so if they try to apply the file as-is, they will get an error (i'll add this to the documentation)

PR #303, opened 2019-08-06
Change configurator version to 2.3.0-1.0.0

bjhaid:


bob-bins:
The version only lives in documentation for now?

bjhaid:
@bob-bins 

> The version only lives in documentation for now?

Yes, when we decide which direction we will go with configurator, hopefully making it live it it's own repo then we can retroactively tag the git sha

PR #305, opened 2019-08-07
Make sysdig super admin user and password configurable.

bjhaid:
This make sysdig.admin.{username,password} configurable, also provides
an helpful message at the end of the script to indicate what the
username and password is and how to login to the UI.


configurator/sysdig-chart/defaultValues.yaml
```diff
@@ -22,13 +22,20 @@ cloudProvider:
   region:
   create_loadbalancer: false
 sysdig:
+  admin:
+    username: test@sysdig.com
+    password:
```
bob-bins: no need for `password` here since it's in the secrets-values.yaml

configurator/sysdig-chart/templates/common-config/config.yaml
```diff
@@ -158,10 +158,10 @@ data:
   # found in the file api-ingress.yaml
   api.url: https://{{ .Values.sysdig.dnsName }}:443
   # Required: Sysdig Cloud super admin user. This will be used for the initial login in the web interface
-  sysdigcloud.default.user: {{ .Values.sysdig.default.username }}
+  sysdigcloud.default.user: {{ .Values.sysdig.admin.username }}
   # Required: Sysdig Cloud super admin user password
   # NOTE: Change upon first login
-  sysdigcloud.default.user.password: {{ .Values.sysdig.default.password }}
+  sysdigcloud.default.user.password: {{ .Values.sysdig.admin.password }}
```
bob-bins: should we start putting passwords in `secret`s instead?

configurator/sysdig-chart/defaultValues.yaml
bjhaid: `yq` will barf when merging YAMLs if `password` is not present

configurator/sysdig-chart/templates/common-config/config.yaml
bjhaid: yes eventually... but that needs to happen in a separate pass

configurator/sysdig-chart/generate_templates.sh
```diff
@@ -41,7 +41,7 @@ if [ -f "$GENERATED_SECRET_FILE" ]; then
     log info "$SECRET_FILE exists"
 else
     log info "Secret file does not exist. Creating Secretfile"
-    helm template -x "templates/$SECRET_FILE" "$TEMPLATE_DIR/secret-generator" > "$GENERATED_SECRET_FILE"
+    helm template -f "$TEMPLATE_DIR/defaultValues.yaml" -x "templates/$SECRET_FILE" "$TEMPLATE_DIR/secret-generator" > "$GENERATED_SECRET_FILE"
```
marojor: @bjhaid  when a custom password is provided in `values.yaml`, the secret file will contain `defaultValues.yaml`, which eventually ends in displaying the wrong password to the user.

Should `values.yaml` be part of this `helm` command to?

bjhaid: Nice catch thanks! I'll PR a fix

PR #306, opened 2019-08-07
Update README to indicated new versioning scheme.

bjhaid:
This also updates the docker tags to match the versioning scheme as
described under the Versioning section.


configurator/README.md
```diff
@@ -45,6 +45,29 @@ The rationale for the workflow is it helps the code reviewers to see the
 effect(s) of the changes introduced, it also helps exercise the config
 generation workflow.
 
+### Versioning
```
bob-bins: Somewhat a separate, but related question - are we going to version the values.yaml(s) as well?

bjhaid: yes it's going to have a schema version, which is completely independent of this version, I am going to open a PR for that

PR #309, opened 2019-08-08
Ensure admin password from values.yaml makes it to secrets-values.yaml.

bjhaid:
The code was using the user provided password in configuring the sysdig
admin password but printing the wrong password at the end of the run to
the user because it was pulling the password from secrets-values.yaml
which had the auto-generated password because values.yaml was not used
in it's templating.

This makes values.yaml considered when generating secret-values.yaml
file, so the user provided `sysdig.admin.password` makes it to
secret-values.yaml.

Credits to @marojor for finding and reporting this.


bjhaid:
This does not have any test code indicating the bugfix because we strip out the password from the generated code:

https://github.com/draios/sysdigcloud-kubernetes/blob/c275c8f83fbf7a36268c19424886557931c8eb4b/configurator/sysdig-chart/test.sh#L9-L20

PR #316, opened 2019-08-23
bump es version

bob-bins:
The change to this `.8` image is that the searchguard assets are downloaded into the image rather than pulled at runtime


bjhaid:
@bob-bins do you mind opening a similar PR for configurator?

PR #318, opened 2019-08-29
Bump redis es

bob-bins:


papajulio:
Do we need this against `onprem-2019-09` branch too? @bob-bins

PR #333, opened 2019-09-19
Remove deprecated APIs because of 1.16 release

figarocorso:
Update YAMLs because of `1.16` release: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
This will make these YAMLs to fail in kubernetes versions prior to `1.9`.

I have not properly tested this, but it can be reviewed; I'd like to test this when `onprem-2019-09` is merged.

### DIFF
--- a/datastores/as_kubernetes_pods/manifests/redis/redis-deployment.yaml
+++ b/datastores/as_kubernetes_pods/manifests/redis/redis-deployment.yaml
@@ -13,11 +13,14 @@ spec:
     app: sysdigcloud
     role: redis
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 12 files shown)


datastores/as_kubernetes_pods/manifests/redis/redis-deployment.yaml
```diff
@@ -13,11 +13,14 @@ spec:
     app: sysdigcloud
     role: redis
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: sysdigcloud-redis
 spec:
+  selector:
```
janoe: It's necessary here the matchSelector
I've tested that is not working with this job:
https://ci.draios.com/view/QA/job/QA-onprem-stack-creator/60/console

PR #337, opened 2019-10-09
SSPROD-2383 : Adding upgrade to anchore 0.5.0 for reports, core

haresh-suresh:
SSPROD-2383 : Adding upgrade to anchore 0.5.0 for reports, core


papajulio:
This has been fixed in `dev` branch we will use the changes from there

PR #338, opened 2019-10-10
Fix anchore upgrade version

haresh-suresh:
Fix anchore upgrade version to have tag name without v prefix


drosen0:
cc @papajulio

PR #339, opened 2019-10-14
WIP: Add k8s policy advisor

mstemm:
Just creating a PR now so it's easier to see the changed files in one place.


mstemm:
We're going to switch to using the sysdig installer, so no need to merge any changes here.

PR #344, opened 2019-12-04
Add k8s policy advisor

mstemm:


sysdigcloud/config.yaml
```diff
@@ -212,6 +212,18 @@ data:
   audit.events.retention.days: "14"
   audit.events.max.events.per.day: "100000"
 
+  ####################################
+  # POLICY ADVISOR
+  ####################################
+  padvisor.api.endpoint: "0.0.0.0"
+  padvisor.api.port: "8176"
+  padvisor.backend.url: "http://sysdigcloud-api:8080"
```
mbreitung: this seems like a value that is duplicated?
https://github.com/draios/sysdigcloud-kubernetes/blob/ecd91590efd2fe5a124d5f2c45aca1ff79168493/sysdigcloud/config.yaml#L178

mstemm: They both serve the same purpose, naming the ultimate backend uri when the scanning/padvisor service needs to do that, but they are different services so I think it's best to have them as separate config options.

mstemm:
I'm gonna rebase to clean up the commits/messages.

PR #349, opened 2019-12-13
Update openshift collector

figarocorso:


sysdigcloud/openshift/openshift-collector-deployment.yaml
```diff
@@ -128,7 +135,7 @@ spec:
             name: sysdigcloud-config
         - name: ssl
           secret:
-            secretName: sysdigcloud-ssl-secret-collector
+            secretName: sysdigcloud-ssl-secret
```
papajulio: Be careful with this change I think it's not needed... Could be?

PR #351, opened 2019-12-13
Disable NATS by default

figarocorso:
As NATS is needed only for Secure related features. We are disabling it by default and adding that info into docs.


mstemm:
You should ask @lorenzo-david, he's using NATS for profiling.

figarocorso:
Closing the PR for now. I'll might reopen this if needed soon.

PR #366, opened 2020-06-12
Update read-me

mathew-fleisch:


mathew-fleisch:
I added a line to the end of the read-me document to trigger the sonar cloud static code analysis tool.

PR #371, opened 2021-06-09
[QA-2735] Prevent get_support_bundle.sh from exiting  if configmap "sysdigcloud-config" doesn't exist

hectorgarzon:
When we use the get_support_bundle.sh script with the namespace sysdig-agent, there is no configmap called "sysdigcloud-config" and therefore when the command `kubectl ${KUBE_OPTS} get configmap sysdigcloud-config -o yaml | grep -v password > ${LOG_DIR}/config.yaml` is run the script exists without writing the tgz file
Adding just a `|| true` this doesn't happen and the tgz file can be written


hectorgarzon:
I close this PR because this is not the right repo. This PR is the right one: https://github.com/draios/sysdig-cloud-scripts/pull/161

